PR #9, opened 2021-11-04
Make memcoin test more automatic with tmux

augustebaum:


nkcr (commented):
Sweet!
I would add some minimal instructions for someone that never used tmux, like how to navigate between panes and how to exit.

nkcr (approved):

Issue #14, opened 2021-11-23
All ballots should have the same size, pre-determined according to the election format


emduc:
If some ballots are encoded with more chunks of 29 bytes than others, then we could possibly map decrypted ballots to users. They should all have the same size, determined before hand, before the ballot is submitted to the chain

nkcr:
To do that we must first support flexible election structures, instead of just a single list of candidates. You can check the format we though of there: https://github.com/dedis/d-voting/blob/main/docs/state%20of%20smart%20contract.md

You can see that the maximum ballot size can be computed with the `Configuration` of the election, since it contains all the questions.

PR #21, opened 2021-11-30
Poll formats and bigger ballots

### LINKED ISSUE #10: Allow ciphertexts to be bigger than 29 bytes
For now ciphertexts are bounded by Kyber to be <= than 29 bytes. It would be better to represent a cipher text by chunks of 29 bytes. To make sure comparisons between encrypted and decrypted ballots is impossible, the size of all ballots should still be the same

emduc:
Adapts the election process to support more elaborate elections with rank, select and text questions. Ballots should all have the same size, defined in the election struct. Ballots also have a new format which requires a specific encoding to be formally described.

Closes #10  and  #14


nkcr:
Nice to see it moving :)
The new format must be described in the `/docs`. As soon as you have some specification feel free to update the doc (create a new file and update `sidebar.md` if you need a new file).

emduc:
The scenario test has been adapted to contain ballots with 2 chunks and a basic election configuration with two subjects with one question each. 
The branch as it is cannot be merged because it uses the Shuffle of Sequence which has not been merged yet into the main branche of Kyber, it's hard coded in go.mod for now.

contracts/evoting/evoting.go
```diff
@@ -601,3 +616,64 @@ func (e evotingCommand) cancelElection(snap store.Snapshot, step execution.Step)
 
 	return nil
 }
+
+// verifyIdentity is a utility function to verify if a public key is associated
+// to a member of the roster or not. Returns no error if it's the case.
```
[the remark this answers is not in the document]
nkcr: I feel it would be better to have a function `isMemberOf(roster authority.Authority, pubKey crypto.Publikey) bool`

contracts/evoting/evoting.go
```diff
@@ -601,3 +616,64 @@ func (e evotingCommand) cancelElection(snap store.Snapshot, step execution.Step)
...
+type SemiRandomStream struct {
+	// Seed is the seed on which should be based our random number generation
+	seed []byte
+
+	stream *rand.Rand
+}
+
+func NewSemiRandomStream(seed []byte) (SemiRandomStream, error) {
+	if len(seed) > 8 {
+		seed = seed[0:8]
```
nkcr, line 654: I'm a little bit concerned that we use only the first 8 bytes. It means that one only need to forge the 8 first bytes of the SHA256 to compromise the random source.

nkcr: Maybe then using the hash of the hash would do the trick.

docs/ballot_encoding.md
```diff
@@ -0,0 +1,67 @@
...
+
+## Answers
+The answers to questions are encoded in the following way:
+
+```
+type:ID:<answers>
+```
+Where we have that:
+* the type can either be "select", "rank" or "text"  
+* ID is the ID of the question
```
nkcr, line 16: but must have a common encoded size, right ?

docs/ballot_encoding.md
```diff
@@ -0,0 +1,67 @@
...
+## Answers
+The answers to questions are encoded in the following way:
+
+```
+type:ID:<answers>
+```
+Where we have that:
+* the type can either be "select", "rank" or "text"  
+* ID is the ID of the question 
+* The 'answers' are seperated by commas
```
nkcr, line 17: I would try to use some sort of metasyntax (https://en.wikipedia.org/wiki/Metasyntax). For example:

```
<type><sep><id><sep><answers>

TYPE = "select"|"text"|"rank"
SEP = ":"
ID = int8
ANSWERS = <answer>[","<answer>]*
ANSWER = <select_answer>|<text_answer>|<rank_answer>
SELECT_ANSWER = int8
...
```

contracts/evoting/evoting.go
nkcr: Forget it, the hash does not depend on the prover since it comes from the shuffled ballots + the electionID

nkcr (changes requested):
Wow, nice work! 💪 
Looks really nice, congratulation for making it so far. It's exciting to see how thinks are coming nicely.
Sorry for not being ready on the Kyber side. I'm doing what I can to bring attention to the pending PR.

nkcr:
@emduc I made an alpha release of Kyber with the sequences shuffling. You can then use `go.dedis.ch/kyber/v3 v3.1.0-alpha`.

contracts/evoting/types/ballots.go
```diff
@@ -0,0 +1,421 @@
...
+			}
+
+			b.SelectResultIDs = append(b.SelectResultIDs, ID(question[1]))
+			b.SelectResult = append(b.SelectResult, make([]bool, 0))
+
+			index := len(b.SelectResult) - 1
+			var selected uint = 0
+
+			for _, selection := range selections {
+				s, err := strconv.ParseBool(selection)
```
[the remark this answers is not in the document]
emduc: In the end it's the same code because ParseBool works for both true/false and 1/0  :D

contracts/evoting/evoting.go
```diff
@@ -436,53 +460,44 @@ func (e evotingCommand) closeElection(snap store.Snapshot, step execution.Step)
 	if len(closeElectionBuf) == 0 {
 		return xerrors.Errorf(errArgNotFound, CloseElectionArg)
 	}
-
```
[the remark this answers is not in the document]
emduc: Oops, they must have been lost when I rolled back this method

contracts/evoting/evoting.go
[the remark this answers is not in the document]
emduc: => Returns NO error if it's the case, ie. the publicKey is associated to a member of the roster. I agree that it should be "nil" instead of no error.

=> isMemberOf is a better name but if the return type is bool, when I find an error I cannot propagate it, would I just return false ?

docs/ballot_encoding.md
emduc: 2 bytes encoded in base64 (4 chars)

nkcr (approved):

PR #23, opened 2021-12-06
new backend for tequila and sign requests

ambor1011:


web/app/src/components/navigation/NavBar.js
```diff
@@ -14,6 +14,21 @@ const NavBar = () => {
         textShadow:'0 0.015em #177368,0 -0.015em #177368,0.01em 0 #177368,-0.01em 0 #177368',
     };
 
+    const [name, setName] = useState('');
+    const [firstname, setFirstname] = useState('');
+    const [sciper, setSciper] = useState(0);
+
+    fetch('/api/getpersonnalinfo')
```
nkcr, line 21: you're already doing fetching the personal info in `App.js` and you should do it only once. When you do it in `App.js` there you could save all the personal info in a state and use it there instead of doing another call.

web/app/src/components/login/Login.js
```diff
@@ -25,8 +26,30 @@ const Login = ({setToken}) => {
         } catch(error){
             setLoginError(error);
             console.log(error);
+        }*/
+
+
+        try{
+            fetch(GET_TEQ_EENDPOINT).then(resp => {
```
nkcr, line 33: Using asyn/await would make things clearer.

web/app/src/components/utils/ExpressEndoints.js
```diff
@@ -0,0 +1 @@
+export const GET_TEQ_EENDPOINT = '/api/getTkKey';
```
nkcr, line 1: what is the `Tk` for in `getTkKey` ?

web/backend/index.js
```diff
@@ -0,0 +1,186 @@
...
+// Serve the static files from the React app
+app.use(express.static(path.join(__dirname, 'client/build')));
+
+//Express-session
+app.set('trust-proxy', 1);
+
+app.use(cookieParser());
+const oneDay = 1000 * 60 * 60 * 24;
+app.use(session({
+    secret: "thisismysecrctekeyfhrgfgrfrty84fwir767",
```
nkcr, line 26: will have to be in a config

web/backend/index.js
```diff
@@ -0,0 +1,186 @@
...
+app.use(express.urlencoded({ extended: true }));
+
+//serving public file
+//app.use(express.static(__dirname));
+
+
+// An api endpoint that returns a short list of items
+app.get('/api/getTkKey', (req,res) => {
+
+    body = 'urlaccess=https://dvoting-dev.dedis.ch/api/control_key\nservice=Evoting\nrequest=name,firstname,email,uniqueid,allunits';
```
nkcr, line 42: The endpoint should be in a config too. Also there should be a way to test it locally with no real Tequilla authentication.

web/backend/index.js
```diff
@@ -0,0 +1,186 @@
...
+
+                req.session.userid = parseInt(sciper);
+                req.session.name = name;
+                req.session.firstname = firstname;
+
+                //res.cookie('sign', sciper, { maxAge: 900000, httpOnly: true }); //TODO change sciper by signed message and maxAge
+                res.redirect('/');
+
+            } else {
+                res.json('c est pas bon');
```
nkcr, line 76: ... ;)

web/backend/index.js
```diff
@@ -0,0 +1,186 @@
...
+        .catch(error => {
+            console.error(error)
+        });
+});
+
+app.get('/api/control_key', (req, res) => {
+    usr_key = req.query.key;
+    body = 'key=' + usr_key;
+
+    axios
```
nkcr, line 59: you're using `fetch` in the frontend so I would stay with that and not add another dependency.

web/backend/index.js
```diff
@@ -0,0 +1,186 @@
...
+        .post('https://tequila.epfl.ch/cgi-bin/tequila/fetchattributes', body)
+        .then(resa => {
```
nkcr, line 61: `resa` ..?

nkcr (changes requested):
Made a quick pass with some comments. In general the code feels far from being ready to be deployed. At this stage I don't want to merge this PR and I'm expecting a significant amount of cleaning, documentation, test, refactoring, etc... before we can think about including that work.

It includes:

- Format properly the source with ESlint
- Remove/clean old parts that are commented / not necessary anymore
- Comment your code, especially the API endpoints
- Add the corresponding documentation in the `/docs` folder: how is the authentication working ? (add a sequence diagram). What is the back-end doing ? How are requests authenticated ? etc... 
- Make the code look like it is not a toy example (proper error message, correct variable names, etc...

Remember that this PR represents most of the work done during this semester. It should justify an 8 credit amount of work, which is not something I can feel right now.

nkcr (approved):

PR #24, opened 2021-12-15
D-voting integration test with smart contract documentation

jbsv:
Added an integration test to create n nodes and v votes.
Running the test currently shows a comparison between the casted votes and the decrypted votes. However, the decrypted votes read in the proof are still empty, so that this test requires further debugging to understand why the decrypted votes are note the same as the casted votes:

```log
Title of the election : Three votes election
ID of the election : 16a3b508398e330a412f03a2d3e1b45c009f4c063b10e91d553836508229a895
Status of the election : 3
Number of decrypted ballots : 0
```


docs/smart_contract.md
```diff
@@ -0,0 +1,104 @@
...
+Key / Value pairs sent in the transaction in order to create an election:
+| | |
+|-|-|
+|"go.dedis.ch/dela.ContractArg"|[]byte(evoting.ContractName)|
+|evoting.CreateElectionArg|createElectionBuf|
+|evoting.CmdArg|[]byte(evoting.CmdCreateElection|
+
+where:
+    
+    evoting.ContractName = "go.dedis.ch/dela.Evoting"
```
nkcr, line 21: We use the \`\`\` syntax for code. You can then specify the language: https://docs.github.com/en/github/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#syntax-highlighting

docs/smart_contract.md
```diff
@@ -0,0 +1,104 @@
+TODO: 
+1. Explain how to sync up the smart contract with the DKG
+2. Explain how to sync up the smart contract with the Shuffle
+3. Explain how security impacts the smart contract (at the decrypt stage)
+
+## Create an election
+This transaction requires the following 3 parameters:
```
nkcr, line 7: A detail, but it is more pleasant if we space the text with empty lines: After/before titles, code blocks, lists, and tables.

test/dvotingdela.go
```diff
@@ -0,0 +1,519 @@
...
+
+	GetAddr() net.Addr
+	GetPublicKey() crypto.PublicKey
+	GetPool() pool.Pool
+	GetAccessStore() accessstore
+	GetTree() hashtree.Tree
+	GetDkg() dkg.DKG
+	GetShuffle() shuffle.Shuffle
+
+	// jean ->
```
nkcr, line 86: Should we remove it ?

test/dvotingdela.go
```diff
@@ -0,0 +1,519 @@
...
+	// manager   *signed.TransactionManager
+	// firstNode dVotingNode
+	// actor     dkg.Actor
+	// signer    crypto.AggregateSigner
+	// <- jean
+}
+
+// dVotingNode represents a Dela node using cosi pbft aimed to execute d-voting
+//
+// - implements dela
```
nkcr, line 97:
```suggestion
// - implements dVotingCosiDela
```

test/dvotingdela.go
```diff
@@ -0,0 +1,519 @@
...
+	accessStore   accessstore
+	tree          hashtree.Tree
+	dkg           *pedersen.Pedersen
+	shuffle       *neff.NeffShuffle
+}
+
+// Creates n dela nodes using tempDir as root to file path and returns an array
+// of nodes or error
+func setupDVotingNodes(t *testing.T, numberOfNodes int, tempDir string) []dVotingCosiDela {
+	var dVotingNodes []dVotingCosiDela = make([]dVotingCosiDela, numberOfNodes)
```
nkcr, line 116: we never use the `var` to initialize. It is used only to declare a variable. There you should just do 

```go
dVotingNodes := make(...
```

test/dvotingdela.go
```diff
@@ -0,0 +1,519 @@
...
+	var delaNodes []dela = make([]dela, numberOfNodes)
+
+	canal := make(chan dVotingCosiDela, numberOfNodes)
+
+	for i := 0; i < numberOfNodes; i++ {
+		filePath := filepath.Join(tempDir, "node", fmt.Sprint(i))
```
nkcr, line 122:
```suggestion
		filePath := filepath.Join(tempDir, "node", strconv.Itoa(i))
```

test/dvotingdela.go
```diff
@@ -0,0 +1,519 @@
...
+		nodePort := 2000 + i
+
+		go newDVotingNode(canal, t, filePath, nodePort)
+		dVotingNodes[i] = <-canal
```
nkcr, line 126: This operation is blocking so you are not gaining anything. What you probably want to do is something like that:

```go
wait := sync.WaitGroup{}
wait.Add(numberOfNodes)

nodes := make(chan dVotingCosiDela, numberOfNodes)

for i := 0; i < numberOfNodes; i++ {
  go func(i int) {
    defer wait.Done()
    nodePort := 2000 + i
    nodes <- newDVotingNode(t, filePath, nodePort)
  } (i)
}

wait.Wait()
close(nodes)

dVotingNodes := make([]dVotingCosiDela, 0, numberOfNodes)

for node := range nodes {
  dVotingNodes = append(dVotingNodes, node)
}
```

test/dvotingdela.go
```diff
@@ -0,0 +1,519 @@
...
+func (c dVotingNode) GetDkg() dkg.DKG {
+	return c.dkg
+}
+
+// GetShuffle implements dVotingNode
+func (c dVotingNode) GetShuffle() shuffle.Shuffle {
+	return c.shuffle
+}
+
+// generator can generate a private key compatible with the x509 certificate.
```
nkcr, line 396:
```suggestion
// certGenerator can generate a private key compatible with the x509 certificate.
```

test/dvotingdela.go
```diff
@@ -0,0 +1,519 @@
...
+}
+
+func newCertGenerator(r io.Reader, c elliptic.Curve) loader.Generator {
+	return certGenerator{
+		random: r,
+		curve:  c,
+	}
+}
+
+// certGenerator implements loader.Generator. It returns the serialized data of
```
nkcr, line 411:
```suggestion
// Generate implements loader.Generator. It returns the serialized data of
```

nkcr (commented):
Intermediary review.

test/dvotingdela.go
jbsv: shouldn't we close(nodes) after the range loop ?

nkcr: No, otherwise it will loop forever. If I remember correctly once we close a channel we can't add new elements to that channel but we can still dry out its content.

test/dvotingdela.go
jbsv: Indeed :-)

test/dvotingdela.go
jbsv: thanks

test/integration_test.go
```diff
@@ -0,0 +1,566 @@
...
+	t.Logf("initializing shuffle")
+	sActor, err := initShuffle(nodes, signer)
+	require.NoError(t, err)
+
+	t.Logf("shuffling")
+	err = sActor.Shuffle(electionID)
+	require.NoError(t, err)
+
+	// ##### DECRYPT BALLOTS #####
+	time.Sleep(time.Second * 3)
```
jbsv, line 102: missing instructions to decrypt the ballots

.vscode/launch.json
```diff
@@ -0,0 +1,18 @@
+{
```
nkcr, line 1: We normally don't commit IDE specific files

test/dvotingdela.go
```diff
@@ -0,0 +1,510 @@
+package integration
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"crypto/elliptic"
+	"crypto/x509"
+	"io"
+	"math/rand"
```
nkcr, line 9: Didn't check, but we should be sure to set a seed so that tests are reproducible

nkcr (commented):
A lightning review ⚡

emduc (commented):
Everything seems in order for me. 
For the decrption, I believe it should simply be a for loop decrypting each ballot 1 by 1, in the same way it's done in evoting/controller/handler.go :: DecryptBallots , as it's done by the proxy in practice.
By the way, it seems unintuitive to wait a constant amount of time between each transaction / use of service, don't we have a way to block / observe ?

emduc:
>  don't we have a way to block / observe ?

my bad, I thought we slept more than we actually do

Issue #17, opened 2021-11-24
Cast an empty ballot makes shuffle impossible


emduc:
There is currently no check on the ballots submitted using the CastVote() command. If a node submits an empty ballot, the problem will be spotted only during the shuffle and the nodes will never be able to proceed.

A unit test has been added in /contracts/evoting/mod_test.go which should fail but currently passes.

To spot the problem, one can replace a casted ballot in the ScenarioTest by a simple CiphterText{}

emduc:
Furthermore: what if the ciphertext appears valid but is not encrypted using the roster's public key?

PR #25, opened 2021-12-21
ensure submitted ballot is correct size and with non empty ElG Pairs

emduc:
Addresses issues #17  to ensure that when a vote is casted, it's non empty and it won't fail the shuffle

### DIFF
--- a/contracts/evoting/mod_test.go
+++ b/contracts/evoting/mod_test.go
@@ -194,6 +194,26 @@ func TestCommand_CastVote(t *testing.T) {
 
 	_ = snap.Set(dummyElectionIdBuff, jsElection)
 	err = cmd.castVote(snap, makeStep(t, CastVoteArg, string(jsCastVoteTransaction)))
+	require.EqualError(t, err, "the ballot has unexpected length: 1 != 0")
+
+	dummyElection.BallotSize = 29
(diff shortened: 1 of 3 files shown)


contracts/evoting/mod_test.go
```diff
@@ -192,6 +192,56 @@ func TestCommand_CastVote(t *testing.T) {
 	dummyElection.Status = types.Open
 	jsElection, _ = json.Marshal(dummyElection)
 
+	_ = snap.Set(dummyElectionIdBuff, jsElection)
```
nkcr, line 195: you should not ignore an error

nkcr: same bellow

nkcr (approved):
Nice 👍 
Just a question: what happens if the ballot is not encrypted with the public key ? I guess it would just end with a garbage decrypted ballot, but until we are sure about that I would create a new issue for this specific case, before closing #17 .

emduc:
> 

You're right, I added a ballot from a different public key in the scenario test and it the decryption seems to never end :(. 
I commented this ballot in the scenario so that the it still works until we fix the issue, should I uncomment it ? Remove it ?

nkcr (approved):
Thank you, works fine like that!

PR #28, opened 2022-01-03
Decrypt the ballots

jbsv:
The integration test was amended to add a step in order to decrypt the ballots after they have been shuffled.

### DIFF
--- a/test/integration_test.go
+++ b/test/integration_test.go
@@ -44,7 +44,7 @@ func TestIntegration_ThreeVotesScenario(t *testing.T) {
 	// ##### SETUP ENV #####
 	// make tests reproducible
 	rand.Seed(1)
-	delaPkg.Logger = zerolog.New(os.Stdout).Level(zerolog.WarnLevel)
+	delaPkg.Logger = zerolog.New(os.Stdout).Level(zerolog.InfoLevel)
 
 	dirPath, err := ioutil.TempDir(os.TempDir(), "d-voting-three-votes")
 	require.NoError(t, err)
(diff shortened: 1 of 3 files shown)


test/integration_test.go
```diff
@@ -545,6 +557,63 @@ func initShuffle(nodes []dVotingCosiDela, signer crypto.AggregateSigner) (shuffl
...
+	}
+
+	args := []txn.Arg{
+		{Key: "go.dedis.ch/dela.ContractArg", Value: []byte(evoting.ContractName)},
+		{Key: evoting.DecryptBallotsArg, Value: decryptBallotsBuf},
+		{Key: evoting.CmdArg, Value: []byte(evoting.CmdDecryptBallots)},
+	}
+	_, err = m.addAndWait(args...)
+	if err != nil {
+		return xerrors.Errorf("failed to Marshall closeElection: %v", err)
```
emduc, line 611: The error message seems off

test/integration_test.go
```diff
@@ -545,6 +557,63 @@ func initShuffle(nodes []dVotingCosiDela, signer crypto.AggregateSigner) (shuffl
...
+	}
+
+	decryptedBallots := make([]types.Ballot, 0, len(election.ShuffleInstances))
+	wrongBallots := 0
+
+	for i := 0; i < len(X[0]); i++ {
+		// decryption of one ballot:
+		marshalledBallot := strings.Builder{}
+		for j := 0; j < len(X); j++ {
+			chunk, err := actor.Decrypt(X[j][i], Y[j][i], []byte(election.ElectionID))
```
emduc, line 577: the electionID should be obtained with hex.DecodeString here

test/integration_test.go
jbsv: completely, indeed

PR #30, opened 2022-01-11
Fixes the integration test

nkcr:
This PR fixes the integration tests, and notably the nonce synchronization issue. It has been solved by using a transaction manager that uses the validation service to get the nonce, which is the "official" way of doing that. Additionally, each node in the integration test uses its own identity to sign transactions, instead of having one shared identity.

### DIFF
--- a/services/shuffle/neff/controller/action.go
+++ b/services/shuffle/neff/controller/action.go
@@ -4,6 +4,10 @@ import (
 	"github.com/dedis/d-voting/services/shuffle"
 	"go.dedis.ch/dela"
 	"go.dedis.ch/dela/cli/node"
+	"go.dedis.ch/dela/core/access"
+	"go.dedis.ch/dela/core/ordering"
+	"go.dedis.ch/dela/core/txn/signed"
+	"go.dedis.ch/dela/core/validation"
 	"golang.org/x/xerrors"
 )
 
@@ -29,7 +33,24 @@ func (a *initAction) Execute(ctx node.Context) error {
 		return xerrors.Errorf("failed to get signer: %v", err)
 	}
 
-	actor, err := neffShuffle.Listen(signer)
+	var service ordering.Service
+	err = ctx.Injector.Resolve(&service)
+	if err != nil {
+		return xerrors.Errorf("failed to resolve ordering.Service: %v", err)
+	}
+
+	var vs validation.Service
+	err = ctx.Injector.Resolve(&vs)
+	if err != nil {
+		return xerrors.Errorf("failed to resolve validation.Service: %v", err)
+	}
+
+	client := client{
+		srvc: service,
+		mgr:  vs,
+	}
+
+	actor, err := neffShuffle.Listen(signed.NewManager(signer, &client))
 
 	if err != nil {
 		return xerrors.Errorf("failed to initialize the neff shuffle	protocol: %v", err)
@@ -38,4 +59,25 @@ func (a *initAction) Execute(ctx node.Context) error {
 	ctx.Injector.Inject(actor)
 	dela.Logger.Info().Msg("The shuffle protocol has been initialized successfully")
 	return nil
+}
+
+// client fetches the last nonce used by the client
+//
+// - implements signed.Client
+type client struct {
+	srvc ordering.Service
+	mgr  validation.Service
+}
+
+// GetNonce implements signed.Client. It uses the validation service to get the
+// last nonce.
+func (c *client) GetNonce(id access.Identity) (uint64, error) {
+	store := c.srvc.GetStore()
+
+	nonce, err := c.mgr.GetNonce(store, id)
+	if err != nil {
+		return 0, xerrors.Errorf("failed to get nonce from validation: %v", err)
+	}
+
+	return nonce, nil
 }
--- a/services/shuffle/neff/controller/action.go
+++ b/services/shuffle/neff/controller/action.go
@@ -66,7 +66,7 @@ func makeClient(ctx node.Context) (client, error) {
 
 	client := client{
 		srvc: service,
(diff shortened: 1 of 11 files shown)


services/shuffle/neff/controller/action.go
```diff
@@ -29,7 +33,24 @@ func (a *initAction) Execute(ctx node.Context) error {
...
+	var vs validation.Service
+	err = ctx.Injector.Resolve(&vs)
+	if err != nil {
+		return xerrors.Errorf("failed to resolve validation.Service: %v", err)
+	}
+
+	client := client{
+		srvc: service,
+		mgr:  vs,
+	}
```
pierluca, line 51: to keep this function short and sweet, I would consider splitting this in a dedicated function `makeClient(ctx node.Context) (client, err)`.

services/shuffle/neff/handler.go
```diff
@@ -29,7 +27,7 @@ import (
 	"golang.org/x/xerrors"
 )
 
-const watchTimeout = time.Second * 2
+const watchTimeout = time.Second * 6
```
pierluca, line 30: is there any justification for 2 or 6 ? 😅

services/shuffle/neff/mod.go
```diff
@@ -192,7 +189,7 @@ func (a *Actor) waitAndCheckShuffling(electionID string) error {
 		}
 
 		dela.Logger.Info().Msgf("waiting a while before checking election: %d", i)
-		time.Sleep(waitT)
+		time.Sleep(time.Second * time.Duration(rosterLen/2+1))
```
pierluca, line 192: Only for my understanding, why should the sleep be proportional to the roster size ?
Isn't the processing time correlated with the amount of votes that need to be shuffled ?

test/integration_test.go
```diff
@@ -233,56 +247,76 @@ func TestIntegration_ManyVotesScenario(t *testing.T) {
...
+						if err != nil {
+							return nil, xerrors.Errorf("failed to sync: %v", err)
+						}
 
-				return sentTxnID, nil
+						break events
+					}
+
+					return sentTxnID, nil
+				}
```
pierluca, line 315: this, on the surface at least, looks like a good candidate for decomposition into separate functions.
In particular I'd consider using a sentinel error instead of a break, I believe this could simplify the code a lot.

pierluca (commented):
I'm (by far) not an expert in this code, but the changes seem pretty neat.
I would just strive to keep function complexity low and decompose further wherever possible.

services/shuffle/neff/mod.go
nkcr: The shuffling computation is one element, the other one is the consensus time and complexity, which grows with the number of servers.

services/shuffle/neff/handler.go
nkcr: A reasonable amount of time after which you can assume Dela won't include your transaction ... 😁
I've relaxed it to accommodate for the integration test. The timing is a real question though, that is far from trivial. There are a lot of factors (number of nodes, network, CPU power, election and consensus parameters, etc...) that should be taken into account to be able to "formally" compute it. Right now we go with the "good enough" solution.

test/integration_test.go
```diff
@@ -202,7 +216,7 @@ func TestIntegration_ManyVotesScenario(t *testing.T) {
 	require.NoError(t, err)
 
 	// ##### DECRYPT BALLOTS #####
-	time.Sleep(time.Second * 1)
+	time.Sleep(time.Second * 10)
```
jbsv, line 219: Normally a computer can crunch tons of processing in a second. It would be interesting to benchmark things.

nkcr: The beauty of consensus... 😁 This is not only purely about executing opcodes, but there are nodes that need to process transaction, wait some time to gather them, check if the leader is acting correctly, propose to elect a new leader, revert back because the leader woke up, etc... All this takes some time, and it is hard to predict it precisely. And the more we add nodes and transaction, the more we increase the incertitude.

test/integration_test.go
```diff
@@ -58,11 +59,16 @@ func TestIntegration_ThreeVotesScenario(t *testing.T) {
 
 	signer := createDVotingAccess(t, nodes, dirPath)
 
-	m := newTxManager(signer, nodes[0], time.Second*10)
+	m := newTxManager(signer, nodes[0], time.Second*10, 10)
 
 	err = grantAccess(m, signer)
 	require.NoError(t, err)
 
+	for _, n := range nodes {
```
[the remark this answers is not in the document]
nkcr: subtle change: I grant access on `n.GetShuffleSigner()`, which means that each node has its own signer for the shuffling service.

test/integration_test.go
```diff
@@ -289,37 +289,37 @@ func (m txManager) addAndWait(args ...txn.Arg) ([]byte, error) {
...
-				if bytes.Equal(sentTxnID, fetchedTxnID) {
-					accepted, status := event.Transactions[0].GetStatus()
+		err = m.m.Sync()
+		if err != nil {
+			return nil, xerrors.Errorf("failed to sync: %v", err)
+		}
 
-					if !accepted {
-						if i+1 == m.retry {
-							return nil, xerrors.Errorf("transaction not accepted: %s", status)
```
pierluca: no need for this timeout/max-retry anymore ?

pierluca (approved):
LGTM :)

jbsv (approved):
LGTM + thanks for all that

test/integration_test.go
nkcr: The mechanism is still there, but thanks to the refactoring no need for this ugly check anymore :)

PR #36, opened 2022-01-17
Custom Performance tests

emduc:
Provides a basic benchmark, customable in the number of nodes, votes and chunks per ballot.

 The votes being casted are the same vote for all users and contains an empty answer to a text question (full base64-padding).

Implementing a Benchmark required to change the signature of a few methods in the framework of the integrationTest so that they take a more generic type as parameter (also accept testing.B instead of just testing.T)

### DIFF
--- a/.github/workflows/go_integration_tests.yml
+++ b/.github/workflows/go_integration_tests.yml
@@ -19,5 +19,5 @@ jobs:
         go-version: ^1.17
 
 
-    - name: Test integration with coverage
-      run: go test -cover ./test/...
+    - name: Test integration & benchmark with coverage
+      run: go test -bench -cover ./test/...
--- a/test/dvotingdela.go
+++ b/test/dvotingdela.go
@@ -11,7 +11,6 @@ import (
 	"path/filepath"
 	"strconv"
 	"sync"
-	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


jbsv (changes requested):
So far looks good, thanks. 
For the long term, can you please add a build in '.github/workflows/go_test.yml' so that this benchmark compiles in CI? The goal is to avoid that any change breaks the benchmark unnoticed.

nkcr:
Sweet! That's going to be useful. 

The code is still using some ugly `time.Sleep` though. This hack is ok for the integration test, but it doesn't make sense for a benchmark. A proper way of replacing those `time.Sleep` would be to watch for the blockchain: instead of waiting some time for the election to move to the next state for example, we should watch for the block that stores this new state.

I'm gonna try to update the PR.

emduc:
> So far looks good, thanks. For the long term, can you please add a build in '.github/workflows/go_test.yml' so that this benchmark compiles in CI? The goal is to avoid that any change breaks the benchmark unnoticed.

I added a -bench flag so that the benchmark is also executed in the CI, is that enough or did you mean that I should create a new dedicated yml file ?

jbsv (approved):

PR #42, opened 2022-01-27
Fixes the memcoin test

nkcr:
- DKG wasn't closing its DB (! @augustebaum )
- At the end it makes much more sense that DKG uses the Dela DB, it solves the problem of injecting multiple instances of the `kv.DB` object.
- Adapts the DKG tests consequently


augustebaum:
Sorry about this! Indeed it's a lot simpler to reuse the Dela DB.

jbsv (approved):
LGTM

PR #43, opened 2022-02-01
Fall semester clean up

nkcr:
The clean up mostly concerned the message formats. Now (nearly) all messages use the [serde](https://dedis.github.io/dela/#/serde) module, which allows to decouple the types to their supported formats. The JSON format has been implemented.

Otherwise a lot of global code clean up: add/remove/update comments, address some TODOs, update some tests, etc.. 

Some other key aspects I sadly didn't have time to do:
- update the HTTP proxy
- update the smart contract to behave more like cothority with "instances" of elections.


contracts/evoting/controller/handlers.go
```diff
@@ -36,27 +38,27 @@ func (h *votingProxy) Login(w http.ResponseWriter, r *http.Request) {
...
+		Configuration: req.Configuration,
+		AdminID:       req.AdminID,
 	}
 
-	payload, err := json.Marshal(createElectionTransaction)
+	data, err := createElection.Serialize(h.context)
 	if err != nil {
 		http.Error(w, "failed to marshal CreateElectionTransaction: "+err.Error(),
 			http.StatusInternalServerError)
-		return
```
emduc, line 56: Seems like a typo, why not return after the error ?

nkcr: Ooop true. Thanks!

contracts/evoting/types/ballots_test.go
```diff
@@ -95,7 +96,7 @@ func TestBallot_Unmarshal(t *testing.T) {
 
 	// with line wrongly formatted
 	err = b.Unmarshal("x", election)
-	require.EqualError(t, err, "a line in the ballot has length != 3")
+	require.EqualError(t, err, "a line in the ballot has length != 3: x")
```
pierluca, line 99: Why add `: x` ?

contracts/evoting/types/ballots_test.go
```diff
@@ -468,5 +469,72 @@ func TestSubject_IsValid(t *testing.T) {
 
 	valid = configuration.IsValid()
 	require.False(t, valid)
+}
 
+func TestBallot_Equal(t *testing.T) {
```
pierluca, line 474: I bet you know what I'm thinking 😅 but yeah, splitting into smaller test execution units would be an improvement.
In particular, this would be a perfect candidate for a table-driven test.

contracts/evoting/mod_test.go
```diff
@@ -3,7 +3,6 @@ package evoting
 // todo: json marshall and unmarshall branch is are not covered yet
```
pierluca: I understand we don't necessarily have the time, but if Emilien could make a pass to split the tests into smaller units, I really believe it would be a significant improvement and it'd give a better example to the next batch of students. 😇

contracts/evoting/types/ballots_test.go
nkcr: I updated the error to return the faulty line.

services/shuffle/neff/mod.go
```diff
@@ -210,16 +209,43 @@ func (a *Actor) Verify(suiteName string, Ks []kyber.Point, Cs []kyber.Point,
...
+	if proof == nil {
+		return election, xerrors.Errorf("election does not exist")
+	}
+
+	fac := ctx.GetFactory(etypes.ElectionKey{})
+	if fac == nil {
+		return election, xerrors.New("election factory not found")
+	}
+
+	message, err := fac.Deserialize(ctx, proof.GetValue())
```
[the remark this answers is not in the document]
nkcr: Yes, Dela always return a proof that has been checked.

PR #47, opened 2022-02-07
New decryption protocol

emduc:
First draft of the new decryption process. I am still trying to debug the submission of pubShares but the basic architecture is implemented.

What's been done: 
 - Now we send all the pubShares of a given node on the chain, through the DKG service.
 -  The decryption is done in two phases, first we "BeginDecryption", which asks all the nodes to send their pubShare, then we send a DecryptBallotsTx which will finish the decryption using the pubShares submitted.
 - Adapted the integration test to reflet the new decryption flow.

What's still to be done:
 - Fix the submission of pubShares (I added Rounds but each submission still seems to overwrite the last submission...)
 - Adapt all the unit tests / scenarioTest
 - Compare performances

Some thoughts of improvements once the system works: (thought of them too late)
 - the "index" in the pubShares seems to be the same for all the shares of a node, we could store it only once...
 - could avoid some code duplication with a "SignedTransaction" interface that would contain the basic primitives to deal with the transactions that need to be signed (Shuffle, RegisterPubShares)


contracts/evoting/evoting.go
```diff
@@ -490,12 +630,44 @@ func (e evotingCommand) decryptBallots(snap store.Snapshot, step execution.Step)
...
+	nbrPairsPerBallot := len(pubSharesSubmitted[0][0])
+
+	for i := 0; i < nbrBallots; i++ {
+		// decryption of one ballot:
+		marshalledBallot := strings.Builder{}
+
+		for j := 0; j < nbrPairsPerBallot; j++ {
+			chunk, err := Decrypt(i, j, pubSharesSubmitted)
+			if err != nil {
+				return xerrors.Errorf("failed to decrypt (K, C) : ", err)
```
nkcr, line 652:
```suggestion
				return xerrors.Errorf("failed to decrypt (K, C): %v", err)
```

contracts/evoting/types/election.go
```diff
@@ -255,6 +259,25 @@ func (p *PublicBulletinBoard) DeleteUser(userID string) bool {
 	return false
 }
 
+// PubShare represents a public share.
+type PubShare struct {
```
nkcr, line 263: I think we can re-work a bit the data structures.
Since every node must publish their pubshares with their given index, we could only have a slice of pubshares, where the index denotes the node's index, which is the same as the index of the pubshare. Like so we also don't need to store the public key because it is already stored in the roster.

services/dkg/pedersen/handler.go
```diff
@@ -127,36 +142,98 @@ mainSwitch:
...
+		for {
+			election, err := getElection(h.context, msg.GetElectionId(), h.service)
+			if err != nil {
+				return xerrors.Errorf("could not get the election: %v", err)
+			}
+
+			//TODO: Works with current "shuffleThreshold", but the shuffle threshold
+			// should be smaller in theory ? (1/3 + 1 vs 2/3 + 1 ? )
+			differentPubShares := len(election.PubSharesArchive.PubSharesSubmissions)
+			if differentPubShares >= election.ShuffleThreshold {
```
nkcr, line 195: yeah, now we should have a generic `securityThreshold` as the election attribute, which denotes the shuffle and decryption threshold.

nkcr (commented):
Thanks! 👍
Made a quick, intermediary, review. I changed the base branch to mine so that we see only your changes. 

I made quite some new changes on my branch in an effort to clean up the code base. Notably the transaction and election formats. Please try to update your branch with my changes, or at least have a look. We can have a look together to see how to update your code to include my changes. Let me know.

contracts/evoting/evoting.go
```diff
@@ -471,6 +474,143 @@ func (e evotingCommand) closeElection(snap store.Snapshot, step execution.Step)
...
+// registerPubShares implements commands.It performs the REGISTER_PUB_SHARES command
+func (e evotingCommand) registerPubShares(snap store.Snapshot, step execution.Step) error {
+	var tx types.RegisterPubSharesTransaction
+
+	err := getTransaction(step.Current, &tx)
+	if err != nil {
+		return xerrors.Errorf(errGetTransaction, err)
+	}
+
+	err = checkPreviousPubSharesTransactions(step, tx.Round)
```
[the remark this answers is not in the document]
emduc: Ok ! I added it because I had some troubles with adding the pubShares, all the node send their tx, all tx are accepted but in the end we only have on set of pubShares...

contracts/evoting/types/election.go
emduc: Regarding the public key, we store it in order to make sure that one node only sends one set of pubShares. Or could we could we do this some other way?

services/dkg/pedersen/handler.go
emduc: Yes, but why do we currently use 2/3 + 1 as a shuffle treshold? If 1/3 + 1 make a shuffle then at least one of the shuffles is correct, right?

contracts/evoting/types/election.go
nkcr: If we assume that we keep ordering in the slice of pubshares according to the node index we know when a node already published its pubshare. For example, with the following slice:

```go
pubshares {
 nil,
 nil,
 {...},
 {...},
 nil,
}
```

We know that nodes 2 and 3 already published their pubshares.

services/dkg/pedersen/handler.go
nkcr: You're right. I need to meditate on that 😄

services/dkg/pedersen/handler.go
```diff
@@ -168,12 +170,10 @@ mainSwitch:
 			publicShares = append(publicShares, ballotShares)
 		}
 
-		h.RUnlock()
```
emduc, line 144: DId you mean to remove the lock altogether ? No we never unlock the readlock from line 155

nkcr: oops, no that's probably an error from my side

nkcr (commented):
Nice, extensive work 👍 

Get a couple of comments, mostly about naming.
Also, as discussed in live:

- adapt the tests
- use the node's signer for DKG
- use a "bag" of pushares, instead of an array

contracts/evoting/evoting.go
```diff
@@ -492,12 +576,45 @@ func (e evotingCommand) decryptBallots(snap store.Snapshot, step execution.Step)
...
+		return xerrors.Errorf("the public shares have not been submitted,"+
+			" current status: %d", election.Status)
+	}
+
+	allPubShares := election.PubShareSubmissions
+
+	nbrShuffles := len(election.ShuffleInstances)
+
+	nbrBallots := len(election.ShuffleInstances[nbrShuffles-1].ShuffledBallots)
+	nbrPairsPerBallot := len(election.ShuffleInstances[nbrShuffles-1].ShuffledBallots[0])
```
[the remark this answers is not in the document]
emduc: What do you mean by using `seqSize`?
It does panic when there are no shuffles but we should not be able to get to the decryption phase without > 1 ballot, and each ballot has >= 1 pairs.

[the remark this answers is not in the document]
nkcr: Ok, in this case it would be helpful to add a small comment explaining that.
For `seqSize` I meant renaming the variable.

contracts/evoting/types/transactions.go
```diff
@@ -193,6 +194,33 @@ func (sb ShuffleBallots) Serialize(ctx serde.Context) ([]byte, error) {
...
+
+	data, err := format.Encode(ctx, rp)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to encode register pubShares: %v", err)
+	}
+
+	return data, nil
+}
+
+// DecryptBallots ...
```
nkcr, line 223: To be removed

contracts/evoting/types/transactions.go
```diff
@@ -250,7 +278,10 @@ func RandomID() (string, error) {
 // Fingerprint implements serde.Fingerprinter. If creates a fingerprint only
 // based on the electionID and the shuffled ballots.
 func (sb ShuffleBallots) Fingerprint(writer io.Writer) error {
-	writer.Write([]byte(sb.ElectionID))
+	_, err := writer.Write([]byte(sb.ElectionID))
+	if err != nil {
+		return xerrors.Errorf("failed to write election ID to fingerprint: %v", err)
```
nkcr, line 283:
```suggestion
		return xerrors.Errorf("failed to write the election ID: %v", err)
```

contracts/evoting/types/transactions.go
```diff
@@ -261,3 +292,23 @@ func (sb ShuffleBallots) Fingerprint(writer io.Writer) error {
 
 	return nil
 }
+
+// Fingerprint implements serde.Fingerprinter
+func (rp RegisterPubShares) Fingerprint(writer io.Writer) error {
+	_, err := writer.Write([]byte(rp.ElectionID))
+	if err != nil {
+		return xerrors.Errorf("failed to write election ID to fingerprint: %v", err)
```
nkcr, line 300:
```suggestion
		return xerrors.Errorf("failed to write the election ID: %v", err)
```

contracts/evoting/types/transactions.go
```diff
@@ -261,3 +292,23 @@ func (sb ShuffleBallots) Fingerprint(writer io.Writer) error {
...
+	}
+
+	_, err = writer.Write([]byte(strconv.Itoa(rp.Index)))
+	if err != nil {
+		return xerrors.Errorf("failed to write pubShare index to fingerprint: %v", err)
```
nkcr, line 305:
```suggestion
		return xerrors.Errorf("failed to write  the pubShare: %v", err)
```

services/dkg/mod.go
```diff
@@ -31,7 +32,10 @@ type Actor interface {
 	GetPublicKey() (kyber.Point, error)
 
 	Encrypt(message []byte) (K, C kyber.Point, remainder []byte, err error)
-	Decrypt(K, C kyber.Point) ([]byte, error)
+
+	// ComputePubshares sends a decryption request to all nodes in order to gather
+	// the public shares
```
nkcr, line 37:
```suggestion
	// ComputePubshares sends a decryption request to all nodes. Nodes will then publish their public shares on the smart contract.
```

services/dkg/pedersen/handler.go
```diff
@@ -417,44 +410,117 @@ func (h *Handler) handleDeal(msg types.Deal, from mino.Address, addrs []mino.Add
...
-		return false, xerrors.Errorf("failed to deserialize election: %v", err)
+		return xerrors.Errorf("failed to sync manager: %v", err)
 	}
 
-	election, ok := message.(etypes.Election)
-	if !ok {
-		return false, xerrors.Errorf("wrong election type: %T", election)
-	}
+	// loop until our transaction has been accepted, or enough nodes
+	// submitted their pubShares
```
nkcr, line 449:
```suggestion
	// loop until our transaction has been accepted, or enough nodes submitted
	// their pubShares
```

services/dkg/pedersen/handler.go
```diff
@@ -703,3 +769,126 @@ func (s *state) UnmarshalJSON(data []byte) error {
 
 	return nil
 }
+
+// watchTx checks the transaction to find one that match txID. Return if the
```
nkcr, line 773:
```suggestion
// watchTx checks the transaction to find one that match txID. Returns if the
```

services/dkg/pedersen/handler_test.go
```diff
@@ -24,21 +30,73 @@ func TestHandler_Stream(t *testing.T) {
...
+		fake.NewRecvMsg(fake.NewAddress(0), types.NewDecryptRequest(electionIDHex)),
+	)
+
+	emptyPubsharesUnits := electionTypes.PubsharesUnits{
+		Pubshares: make([]electionTypes.PubsharesUnit, 0),
+		PubKeys:   make([][]byte, 0),
+		Indexes:   make([]int, 0),
+	}
+
+	fakeElection := electionTypes.Election{
```
nkcr, line 58:
```suggestion
	election := electionTypes.Election{
```

services/dkg/pedersen/handler_test.go
```diff
@@ -24,21 +30,73 @@ func TestHandler_Stream(t *testing.T) {
...
+	require.EqualError(t, err, "could not send pubShares: failed to check"+
+		" if the shuffle is over: could not get the election: election does not exist: <nil>")
+
+	electionIDHex := hex.EncodeToString([]byte("fakeElection"))
+
+	receiver = fake.NewReceiver(
+		fake.NewRecvMsg(fake.NewAddress(0), types.NewDecryptRequest(electionIDHex)),
+	)
+
+	emptyPubsharesUnits := electionTypes.PubsharesUnits{
```
nkcr, line 52:
```suggestion
	units := electionTypes.PubsharesUnits{
```

services/dkg/pedersen/handler_test.go
```diff
@@ -187,6 +245,102 @@ func TestState_MarshalJSON(t *testing.T) {
 	requireStatesEqual(t, s1, s2)
 }
 
+func TestHandler_HandlerDecryptRequest(t *testing.T) {
+	electionIDHex := hex.EncodeToString([]byte("fakeElection"))
+
+	emptyPubsharesUnits := electionTypes.PubsharesUnits{
```
nkcr, line 251:
```suggestion
	units := electionTypes.PubsharesUnits{
```

services/dkg/pedersen/handler_test.go
```diff
@@ -187,6 +245,102 @@ func TestState_MarshalJSON(t *testing.T) {
...
+		Pubshares: make([]electionTypes.PubsharesUnit, 0),
+		PubKeys:   make([][]byte, 0),
+		Indexes:   make([]int, 0),
+	}
+
+	fakeElection := electionTypes.Election{
```
nkcr, line 257:
```suggestion
	election := electionTypes.Election{
```

services/dkg/pedersen/mod.go
```diff
@@ -56,32 +59,36 @@ const (
...
-	factory     serde.Factory
-	service     ordering.Service
-	electionFac serde.Factory
-	actors      map[string]dkg.Actor
+	mino            mino.Mino
+	factory         serde.Factory
+	service         ordering.Service
+	electionFac     serde.Factory
+	pool            pool.Pool
+	pubSharesSigner crypto.Signer
```
nkcr, line 67:
```suggestion
	signer crypto.Signer
```

services/dkg/pedersen/mod.go
```diff
@@ -56,32 +59,36 @@ const (
...
+	actors          map[string]dkg.Actor
 }
 
 // NewPedersen returns a new DKG Pedersen factory
-func NewPedersen(m mino.Mino, service ordering.Service,
-	electionFac serde.Factory) *Pedersen {
+func NewPedersen(m mino.Mino, service ordering.Service, pool pool.Pool,
+	electionFac serde.Factory, pubSharesSigner crypto.Signer) *Pedersen {
```
nkcr, line 73:
```suggestion
	electionFac serde.Factory, signer crypto.Signer) *Pedersen {
```

services/dkg/pedersen/mod.go
```diff
@@ -95,20 +102,29 @@ func (s *Pedersen) Listen(electionIDBuf []byte) (dkg.Actor, error) {
 		return actor, xerrors.Errorf("actor already exists for electionID %s", electionID)
 	}
 
-	return s.NewActor(electionIDBuf, NewHandlerData())
+	err := txmngr.Sync()
```
nkcr, line 105: Are you sure we need to sync there ?

nkcr (commented):
Close to be mergeable :)

services/dkg/pedersen/mod.go
emduc: Not anymore you're right, it was for the scenario test but in the end I just synced earlier in the handler instead

nkcr (approved):
R4M :) 🚀

PR #48, opened 2022-02-15
Backend fixes

pierluca:


nkcr (approved):

web/backend/nodemon.json
```diff
@@ -0,0 +1,6 @@
+{
```
nkcr, line 1: If we can keep the number of configuration files smaller than the number of typescript files that would be good 😛.
Here I would add that configuration to the package.json: https://github.com/remy/nodemon#packagejson

jbsv (approved):

PR #50, opened 2022-02-18
Backend mocks

jbsv:
Mocked some backend functionalities such as:
- login
- logout
- fetching list of elections
- election creation

Students no longer need to start the real backend to develop the frontend.

This mocked backend also documents the API to the backend.


pierluca (approved):
Approving semi-blindly

PR #52, opened 2022-02-22
Fixes login with Tequila

nkcr:
In short the proxy wasn't working because the extension to `setupProxy` must be `.js`.

- adds 404 message
- applies automatic formatting on some files
- fixes the login callback with tequila
- enables HTTPS by default
- fixes typo "personnal_info"
- adds a conditional import of mocks
- fixes `setupProxy.js` that can't be anything else than a `.js`
- sets strict to false, we are not there yet
- updates the README.md


web/frontend/package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "scripts": {
-    "start": "HTTPS=false react-scripts start",
+    "start": "HTTPS=true react-scripts start",
```
jbsv, line 6: does it still work with the mocked backend ?

web/frontend/src/pages/Login.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FC, useState } from 'react';
+import React, { FC, useEffect, useState } from 'react';
```
jbsv, line 1: useEffect not needed

jbsv (approved):
well done for the proxy (and the rest). 
minor things maybe worth looking at before merging

web/frontend/package.json
nkcr: yes, I tried both and it works

PR #53, opened 2022-02-22
Updates the routing

nkcr:
Updates the routing definition with authenticated routes.

- updates the header to have the login/logout at the right place
- minor i18n text update to include to route origin when redirecting to the login page
- add a key deleted file in old/

Screenshot 2022-02-22 at 14 38 15 
Screenshot 2022-02-22 at 14 38 00
Screenshot 2022-02-22 at 14 37 52

### DIFF
--- a/web/frontend/src/App.tsx
+++ b/web/frontend/src/App.tsx
@@ -1,5 +1,5 @@
 import React, { FC, Suspense, useEffect, useState } from 'react';
-import { BrowserRouter as Router, Routes, Route } from 'react-router-dom';
+import { BrowserRouter as Router, Routes, Route, Navigate, useLocation } from 'react-router-dom';
 
 import {
   ROUTE_HOME,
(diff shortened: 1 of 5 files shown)


web/frontend/src/App.tsx
```diff
@@ -63,32 +75,34 @@ const App: FC = () => {
...
+                  <RequireAuth>
+                    <Admin />
+                  </RequireAuth>
+                }
+              />
+              <Route path={ROUTE_ABOUT} element={<About />} />
+              <Route path={ROUTE_ELECTION_INDEX} element={<ElectionIndex />} />
+              <Route path={ROUTE_LOGIN} element={<Login />} />
+              <Route path="/" element={<Home />} />
+              <Route path="*" element={<NotFound />} />
```
jbsv, line 104: Maybe we should use a dedicated Page404.tsx (under pages/)

jbsv (approved):

web/frontend/src/App.tsx
nkcr: Yes, that's something to eventually add. Even better would be to have a `Error` component that can render different kind of errors.

PR #63, opened 2022-03-09
Adds prometheus metrics

nkcr:
Depends on https://github.com/dedis/dela/pull/209

For now we mainly have Dela metrics. Additional metrics specific to the d-Voting system will be added later.


metrics/controller/action.go
```diff
@@ -0,0 +1,159 @@
...
+		case err := <-errc:
+			return xerrors.Errorf("failed to listen: %v", err)
+		case <-time.After(waitRetry):
+		}
+	}
+
+	srvAddr := srv.getAddr()
+	if srvAddr == nil {
+		return xerrors.Errorf("failed to start metric server")
+	}
```
pierluca, line 76: According to [this](https://stackoverflow.com/a/32742904), the polling seems entirely avoidable.

pierluca (changes requested):
A few minor comments, I only feel strongly about the makefile 😅 looks pretty neat overalls

metrics/controller/action.go
nkcr: That doesn't solve our problem because we are using an external function to create and start the HTTP server. And we don't want the function to panic with a `log.Fatal(server.Serve(ln)) `.

pierluca (approved):

README.md
```diff
@@ -158,3 +164,24 @@ LLVL=info memcoin --config /tmp/node1 e-voting scenarioTest
...
+
+Build info can be added to the binary with the `ldflags`, at build time. Infos
+are stored on variables in the root `mod.go`. For example:
+
+```sh
+versionFlag="github.com/dedis/d-voting.Version=`git describe --tags`"
+timeFlag="github.com/dedis/d-voting.BuildTime=`date +'%d/%m/%y_%H:%M'`"
+
+go build -ldflags="-X $versionFlag -X $timeFlag" ./cli/memcoin
+```
```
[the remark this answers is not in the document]
nkcr: True, that's explained the line bellow 😄 . I found it interesting to explain a bit the magic behind build-time variables.

nkcr:
Waiting on https://github.com/dedis/dela/pull/209 before merging it.

PR #64, opened 2022-03-15
Packaging

jbsv:
The new scripts provided allow to install and start the unikernel service and then the voting node.


contracts/evoting/unikernel/apps/combine_shares/config.env
```diff
@@ -0,0 +1,10 @@
+# dela config
+export dela_bin=".../cli/memcoin"
+export dela_data=".../tmp/dela"
+export dela_port="9000"
+
+# unikernel config
+export fs_mount="mnt/"
+export kvm_image="./combine_shares_kvm-x86_64"
+export qemu_script="./qemu-guest"
+export network_config="./unikernel_network"
```
jbsv, line 10:
```suggestion
export network_config="./network.env"
```

deb-package/pkg/opt/dedis/dvoting/bin/start-dvoting
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+sleep 5
+
+paths=$1
```
jbsv, line 5:
```suggestion
config_file=$1
```

deb-package/pkg/opt/dedis/dvoting/bin/start-dvoting
```diff
@@ -0,0 +1,10 @@
...
+
+source ${paths}
```
jbsv, line 7:
```suggestion
source ${config_file}
```

deb-package/pkg/var/opt/dedis/dvoting/data/mnt/.keep
```diff
@@ -0,0 +1 @@
+# document how to configure unikernel
```
jbsv, line 1: remove

deb-package/pkg/var/opt/dedis/dvoting/data/dela/root_id.conf
```diff
@@ -0,0 +1 @@
+# document how to configure unikernel
```
jbsv, line 1: TODO: add root ID

deb-package/pkg/lib/systemd/system/unikernel.service
```diff
@@ -0,0 +1,24 @@
...
+
+[Service]
+User=dvoting
+
+StandardInput=tty
+StandardOutput=tty
+TTYPath=/dev/tty10
+
+ExecStartPre=/bin/bash -c "/opt/dedis/dvoting/bin/create-bridge /etc/dedis/dvoting/config.env"
+ExecStart=/bin/bash -c "/opt/dedis/dvoting/bin/start-unikernel /etc/dedis/dvoting/config.env"
```
pierluca, line 14: This makes me uncomfortable: it seems very brittle to me.

1) the deployment paths are now read from a configuration file, which means we can't update them later on.
We've just prevented ourselves from moving things with an update
At the very least, I would move the paths in a directory other than /etc/ , which FPM will associate with a configuration file.
Even better would be to hardcode the paths and not pass them as parameters.

2) normally, shell scripts can be run directly by systemd (e.g. https://tecadmin.net/run-shell-script-as-systemd-service/ ). Why do we need to run them in a bash subprocess ?

deb-package/build-deb.sh
```diff
@@ -0,0 +1,69 @@
...
+# fpm needs an existing output directory
+OUTPUT_DIR="dist"
+mkdir -p $OUTPUT_DIR
+
+fpm \
+    --force -t deb -a all -s dir -C deb -n dedis-dvoting -v ${VERSION} \
+    --iteration ${ITERATION} \
+    --deb-user dvoting \
+    --deb-group dvoting \
+    --depends bridge-utils \
```
pierluca, line 50: we also depend on `bash` being present. 
Unless we don't *really* need bash and we can just rely on `/bin/sh` in our scripts ?

deb-package/build-deb.sh
```diff
@@ -0,0 +1,69 @@
...
+    --depends fuse \
+    --depends qemu-kvm \
+    --depends libvirt-daemon-system \
+    --depends net-tools \
+    --depends sgabios \
+    --depends socat \
+    --depends uuid-runtime \
+    --depends virtinst \
+    --depends virt-manager \
```
pierluca, line 59: do we really need `virt-manager` ? 🤔

deb-package/pkg/after-install.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/sh
+
+# fix permissions
+# dvoting:dedis will be applied automatically on sub dirs
+chown root:root /opt/dedis
+
+# allow ls in sub dirs
+chmod 755 /opt/dedis
+
+# not recursive for nows
```
pierluca, line 10:
```suggestion
# ensure we don't change owner of /lib/systemd/system
```

deb-package/pkg/etc/dedis/dvoting/config.env
```diff
@@ -0,0 +1,12 @@
+# dela config
+export dela_bin="/opt/dedis/dvoting/bin/memcoin"
+export dela_data="/var/opt/dedis/dvoting/data/dela"
+export dela_port="9000"
+export prometheus_addr="127.0.0.1:9100"
+export proxy_addr="127.0.0.1:9080"
```
pierluca, line 6: nothing in this file seems to belong to `/etc`, except these 3 lines.
Keep in mind, `/etc` files do not get automatically updated when updating the package, as they're automatically marked as config files and considered user-modifiable.

deb-package/pkg/lib/systemd/system/unikernel.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
+Description=Unikernel Service
+After=network.target
+Before=dvoting.service
```
pierluca, line 4: Since `dvoting.service` requires this service, I assume we don't need to have this here.

deb-package/pkg/lib/systemd/system/dvoting.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=D-Voting Node Service
+After=network.target
+Requires=unikernel.service
+
+[Service]
+User=dvoting
+
+ExecStartPre=/bin/bash -c "rm -f /var/opt/dedis/dvoting/data/dela/daemon.sock"
+ExecStart=/bin/bash -c "/opt/dedis/dvoting/bin/start-dvoting /etc/dedis/dvoting/config.env"
```
pierluca, line 10: same concerns as for `unikernel.service`

deb-package/pkg/lib/systemd/system/dvoting.service
```diff
@@ -0,0 +1,16 @@
+[Unit]
+Description=D-Voting Node Service
+After=network.target
+Requires=unikernel.service
+
+[Service]
+User=dvoting
+
+ExecStartPre=/bin/bash -c "rm -f /var/opt/dedis/dvoting/data/dela/daemon.sock"
```
pierluca, line 9: this seems hacky, but I guess we can live with it for now 🤷 
Just for future reference, the cleaner way would be to rely on `/run` or `/var/run` .

deb-package/pkg/lib/systemd/system/unikernel.service
```diff
@@ -0,0 +1,24 @@
+[Unit]
```
pierluca, line 1: I'm thinking it's probably best if this service has a less generic name, such as `dvoting-uk.service`, to avoid conflicting with possibly other things running on our partners' VMs.

contracts/evoting/unikernel/apps/combine_shares/start-unikernel
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+paths=$1
+
+source ${paths}
+source ${network_config}
+
+echo "Starting KVM image connected to bridge interface ${bridge_iface} ..."
+"${qemu_script}" -k "${kvm_image}" -e "${fs_mount}" -b "${bridge_iface}" -m 1024 -a "netdev.ipv4_addr=${uk_ip} netdev.ipv4_gw_addr=${uk_gw} netdev.ipv4_subnet_mask=255.255.255.0 -- ${uk_port}"
```
pierluca, line 9: I'd consider putting `1024` in a variable for extra clarity 😄

deb-package/pkg/etc/dedis/dvoting/config.env
pierluca, line 6: I believe these 3 lines ensure that prometheus and the proxy will *not* be accessible remotely, as they're binding to the loopback address. We should probably bind to `0.0.0.0:9100` and `0.0.0.0:9080` respectively.

pierluca, line 6: should we simply move these 3 lines into `network.env` ?

pierluca, line 6: I believe these 2 lines ensure that prometheus and the proxy will *not* be accessible remotely, as they're binding to the loopback address. We should probably bind to `0.0.0.0:9100` and `0.0.0.0:9080` respectively. Am I mistaken?

pierluca (changes requested):
I can only imagine the amount of work that went into this, good job !
I do feel that there are still a few rough edges worth cleaning up before the first release though.

deb-package/pkg/etc/dedis/dvoting/config.env
nkcr: done with the latest changes

deb-package/pkg/opt/dedis/dvoting/bin/start-dvoting
```diff
@@ -6,4 +6,4 @@ paths=$1
 
 source ${paths}
 
-LLVL=info ${dela_bin} --config ${dela_data} start --postinstall --promaddr ${prometheus_addr} --proxyaddr ${proxy_addr} --port ${dela_port}
+LLVL=info ${dela_bin} --config ${dela_data} start --unikerneladdr "172.44.0.2:1024" --unikernelfolder ${fs_mount} --externalip --postinstall --promaddr ${prometheus_addr} --proxyaddr ${proxy_addr} --port ${dela_port}
```
pierluca, line 9: I'd try to get the unikernel address:port from a config file

contracts/evoting/evoting.go
```diff
@@ -643,17 +643,20 @@ func (e evotingCommand) combineShares(snap store.Snapshot, step execution.Step)
...
+	// we must pass to the Unikernel the RELATIVE folder path from the one
+	// mounted, i.e. e.exportFolder. This folder MUST be uniq for each
+	// transaction.
+	txIDHex := hex.EncodeToString(step.Current.GetID())
+	tmp := fmt.Sprintf("decrypted-ballots-%s-%s-%d/", tx.ElectionID[:4], txIDHex[:4], time.Now().UnixNano())
 
-	err = exportPubshares(tmp, election.PubsharesUnits, numBallots, numChunks)
+	dela.Logger.Info().Msgf("using the following folder: %s", tmp)
+
+	err = exportPubshares(filepath.Join(e.exportFolder, tmp), election.PubsharesUnits, numBallots, numChunks)
```
pierluca, line 654: `filepath.Join(e.exportFolder, tmp)` is recomputed multiple times.
Consider using a variable.

contracts/evoting/evoting.go
```diff
@@ -643,17 +643,20 @@ func (e evotingCommand) combineShares(snap store.Snapshot, step execution.Step)
...
-		return xerrors.Errorf("failed to create temp dir: %v", err)
-	}
+	// we must pass to the Unikernel the RELATIVE folder path from the one
+	// mounted, i.e. e.exportFolder. This folder MUST be uniq for each
+	// transaction.
+	txIDHex := hex.EncodeToString(step.Current.GetID())
+	tmp := fmt.Sprintf("decrypted-ballots-%s-%s-%d/", tx.ElectionID[:4], txIDHex[:4], time.Now().UnixNano())
 
-	err = exportPubshares(tmp, election.PubsharesUnits, numBallots, numChunks)
+	dela.Logger.Info().Msgf("using the following folder: %s", tmp)
```
pierluca, line 652: is `Info()` logged by default in our setup?

cli/minocontroller/actions.go
```diff
@@ -0,0 +1,102 @@
+// This file contains the implementation of the controller actions.
+//
+// Documentation Last Review: 07.10.2020
```
pierluca, line 3: Is this file copy-pasted from some other repo ?

cli/minocontroller/mod.go
```diff
@@ -0,0 +1,275 @@
+// Package controller implements a controller for minogrpc.
+//
+// The controller can be used in a CLI to inject a dependency for Mino. It will
+// start the overlay on the start command, and make sure resources are cleaned
+// when the CLI daemon is stopping.
+//
+// Documentation Last Review: 07.10.2020
```
pierluca, line 7: same question

cli/minocontroller/actions.go
pierluca: and if so, is there some reason we can't just use the default implementation of mino for this, if there's one ?

contracts/evoting/evoting.go
nkcr: Yes. (the `LLVL=info` from `deb-package/pkg/opt/dedis/dvoting/bin/start-dvoting`)

cli/minocontroller/actions.go
nkcr: Yes. And yes there is one. The mino controller from Dela would always connect to `127.0.01`, where here we are trying to connect to the public ip address.

README.md
```diff
@@ -184,4 +184,8 @@ timeFlag="github.com/dedis/d-voting.BuildTime=`date +'%d/%m/%y_%H:%M'`"
 go build -ldflags="-X $versionFlag -X $timeFlag" ./cli/memcoin
 ```
 
-Note that `make build` will do that for you.
\ No newline at end of file
+Note that `make build` will do that for you. You can also cross-compile with:
+
+```sh
+GOOS=linux GOARCH=amd64 make build
```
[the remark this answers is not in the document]
nkcr: yes, would be a good idea.

deb-package/pkg/lib/systemd/system/unikernel.service
```diff
@@ -0,0 +1,24 @@
...
+User=dvoting
+
+StandardInput=tty
+StandardOutput=tty
+TTYPath=/dev/tty10
+
+ExecStartPre=/bin/bash -c "/opt/dedis/dvoting/bin/create-bridge /etc/dedis/dvoting/config.env"
+ExecStart=/bin/bash -c "/opt/dedis/dvoting/bin/start-unikernel /etc/dedis/dvoting/config.env"
+
+ExecStop=/bin/kill --signal INT $MAINPID
```
jbsv, line 16: note: just keep INT for QEMU

contracts/evoting/unikernel/apps/config/config.env
```diff
@@ -1,10 +1,11 @@
 # dela config
-export dela_bin=".../cli/memcoin"
-export dela_data=".../tmp/dela"
+export dela_bin="../../../../../memcoin"
+export dela_data="/tmp/dela"
```
pierluca, line 3: are you sure about this ?

nkcr (approved):
Go! 🚀

PR #65, opened 2022-03-17
Updates memcoin to have automatic setup

nkcr:
- with the `--postinstall` option when running `memcoin start`, the CLI will automatically start and initializes its services
- also fixes the Neff handler to sync the manager in case it can't send the transaction to the pool


nkcr:
@pierluca @jbsv I'm merging as is. Can be review in postmortem.

PR #71, opened 2022-03-23
Modify cast ballot to handle the new election format

cmsigrist:
- [x] Display the content of an election in the Vote tab according to the new election format
- [x] Add input checks on the content of a ballot before it can be cast
- [x] Encode ballots
- [x] Add a mock to cast a ballot
- [ ] Make Regex checks and error message more user friendly


web/frontend/src/pages/ballot/ShowTexts.tsx
```diff
@@ -0,0 +1,75 @@
...
+    for (const answer of newAnswers.TextAnswers[questionIndex].Answers) {
+      if (!regexp.test(answer) && answer !== '') {
+        newAnswers.Errors[errorIndex].Message = t('regexpCheck', { regexp: question.Regex });
+      }
+    }
+  }
+  setAnswers(newAnswers);
+};
+
+const textDisplay = (
```
nkcr, line 53: This is wrong. Your [hook](https://reactjs.org/docs/hooks-intro.html) must start with a capital letter, and embed its functions. It should look like as follow:

```ts
type TextDisplayProps = {
  choice: string;
  question: Text;
  answers: Answers;
  setAnswers: React.Dispatch<React.SetStateAction<Answers>>;
};

const TextDisplay: FC<TextDisplayProps> = ({ choice, question, answers, setAnswers }) => {
  const handleTextInput = (e: React.ChangeEvent<HTMLInputElement>) => {
    const { questionIndex, choiceIndex, errorIndex, newAnswers } = getIndices(
      question,
      choice,
      answers,
      TEXT
    );

    newAnswers.TextAnswers[questionIndex].Answers[choiceIndex] = e.target.value.trim();
    newAnswers.Errors[errorIndex].Message = '';

    if (question.Regex) {
      const regexp = new RegExp(question.Regex);
      for (const answer of newAnswers.TextAnswers[questionIndex].Answers) {
        if (!regexp.test(answer) && answer !== '') {
          newAnswers.Errors[errorIndex].Message = t('regexpCheck', { regexp: question.Regex });
        }
      }
    }
    setAnswers(newAnswers);
  };

  return (
    <div>
      <label htmlFor={choice} className="text-gray-600">
        {choice + ': '}
      </label>
      <input
        id={choice}
        type="text"
        key={choice}
        className="mt-1 sm:text-sm border rounded-md w-1/2"
        onChange={handleTextInput}
      />
    </div>
  );
};
```

And used like that:

```ts
                  <div className="pl-8">
                    {(question.Content as Text).Choices.map((choice) => (
                      <TextDisplay
                        choice={choice}
                        question={question.Content as Text}
                        answers={answers}
                        setAnswers={setAnswers}
                      />
                    ))}
                  </div>
```

nkcr: Same for other places. Use Hooks instead of standard functions. A Hook defines a self-contained element, with its functions, states, and view.

nkcr (changes requested):
Thanks, the results looks really nice 👌 

However there seems to be some misuse of React component that requires some rework. Also added some comments about formatting: in general let your code breath with empty returns, instead of concatenating everything. 

I'm waiting for you to update the code with my comments before going deeper into the review.

web/frontend/src/pages/ballot/ShowTexts.tsx
cmsigrist: Ok, yes, it looks better, thank you :)

web/frontend/src/pages/ballot/Show.tsx
```diff
@@ -80,60 +104,144 @@ const Ballot: FC = () => {
...
+          <div className="px-8">
+            <div>
+              {question.Type === SUBJECT ? (
+                <h3 className="font-bold text-lg text-gray-600">{question.Content.Title}</h3>
+              ) : (
+                <h3 className="text-lg text-gray-600">{question.Content.Title}</h3>
+              )}
+            </div>
+            <div>
+              {question.Type === SELECT ? (
```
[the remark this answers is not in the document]
cmsigrist: So I could move the `switch(question.Type)` in a `render` function, which would then return the correct sub-components (`TextDisplay`, etc...) ?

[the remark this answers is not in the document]
nkcr: I was more thinking about updating the `question` interface with `renderText()` `renderSelect()`.... functions.
Then you could have:

```js
{question.Type === SUBJECT ? (
  {question.renderSubject(...)}
)}

{question.Type === TEXT ? (
  {question.renderText(...)}
)}

...
```

web/frontend/src/pages/ballot/ShowRanks.tsx
```diff
@@ -0,0 +1,123 @@
...
+const rankDisplay = (
+  rankIndex: number,
+  choice: string,
+  question: Rank,
+  answers: Answers,
+  setAnswers: React.Dispatch<React.SetStateAction<Answers>>
+) => {
+  return (
+    <Draggable key={choice} draggableId={choice} index={rankIndex}>
+      {(provided) => (
```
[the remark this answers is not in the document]
cmsigrist: To be honest, I'm not entirely sure how this whole drag and drop works... But apparently a Draggable props has children of type 
```
DraggableChildrenFn = (
  **DraggableProvided**,
  DraggableStateSnapshot,
  DraggableRubric,
) => Node; 
```
So I guess this is where the `provided` comes from ?
(https://github.com/atlassian/react-beautiful-dnd/blob/master/docs/api/draggable.md#draggable-props)

web/frontend/package.json
```diff
@@ -26,12 +27,14 @@
     "i18next-browser-languagedetector": "^6.1.3",
     "prop-types": "^15.8.1",
     "react": "^17.0.2",
+    "react-beautiful-dnd": "^13.1.0",
```
nkcr, line 30: You need to add the corresponding typescript types: https://www.npmjs.com/package/@types/react-beautiful-dnd

nkcr: (as a dev dependency: `npm install -D @types/react-beautiful-dnd`)

web/frontend/src/pages/ballot/QuestionDisplay.tsx
```diff
@@ -0,0 +1,107 @@
...
+        <Droppable droppableId={String(rank.ID)}>
+          {(provided) => (
+            <ul
+              className={question.Content.ID}
+              {...provided.droppableProps}
+              ref={provided.innerRef}>
+              {Array.from(
+                answers.RankAnswers.find((r: RankAnswer) => r.ID === rank.ID).Answers.entries()
+              ).map(([rankIndex, choiceIndex]) => (
+                <RankDisplay
```
nkcr, line 40: Be careful when using lists and maps: https://reactjs.org/docs/lists-and-keys.html#keys

web/frontend/src/pages/ballot/ShowRanks.tsx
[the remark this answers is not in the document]
nkcr: As I was curious I discovered this is a "[render-props](https://reactjs.org/docs/render-props.html)". A way for the component to let the user define how the component can be rendered. The component provides the variables (`provided`) and the user defines what to render with them.

This is nothing but another variables passed to the component (i.e. a "render" function). The following is equivalent:

```ts
        <Droppable
          droppableId={String(rank.ID)}
          children={(provided) => (
            <ul
              className={question.Content.ID}
              {...provided.droppableProps}
              ref={provided.innerRef}>
              {Array.from(
                answers.RankAnswers.find((r: RankAnswer) => r.ID === rank.ID).Answers.entries()
              ).map(([rankIndex, choiceIndex]) => (
                <RankDisplay
                  rankIndex={rankIndex}
                  choice={rank.Choices[choiceIndex]}
                  question={rank}
                  answers={answers}
                  setAnswers={setAnswers}
                />
              ))}
              {provided.placeholder}
            </ul>
          )}
        />
```

web/frontend/src/components/utils/useConfiguration.tsx
```diff
@@ -0,0 +1,165 @@
...
+        Message: '',
+      });
+    }
+  }
+
+  return numberOfQuestion;
+}
+
+// On Configuration changes, iterate over the top-level Subjects of the Configuration
+// to sort the questions and initialize the corresponding Answers.
```
nkcr, line 130:
```ts
// On Configuration changes, iterate over the top-level Subjects of the
// Configuration to sort the questions and initialize the corresponding Answers.
```

web/frontend/src/components/utils/useConfiguration.tsx
```diff
@@ -0,0 +1,165 @@
...
+  TextAnswer,
+} from 'types/configuration';
+
+// Flattens and re-orders the questions (as the first object encountered in a
+// Subject might not be the first that needs to be displayed according to the Order).
+// Adds a ParentID for later rendering and initializes the Answers.
+function flattenSubject(
+  subject: Subject,
+  parentId: ID,
+  sortedQuestions: Array<Question>,
```
nkcr, line 26: this array is not sorted, but contains elements that have a sorted "Order" property. Am I correct ? If that's the case then the naming is misleading.

web/frontend/src/pages/ballot/Show.tsx
```diff
@@ -80,76 +92,90 @@ const Ballot: FC = () => {
...
-    sendBallot();
-    userErrors.noCandidate = '';
-    setUserErrors(userErrors);
   };
 
-  const electionClosedDisplay = () => {
-    return <div> {t('voteImpossible')}</div>;
-  };
+  const subjectTree = (sorted: Question[], parentId: ID) => {
+    const questions = sorted.filter((question) => question.ParentID === parentId);
```
nkcr, line 104: That's pretty inefficient. Why do you need an ordered "flat" list of elements after all ? 
Each subject should have: 

1. The list of element IDs, to specify the subject elements' order
2. A dictionary that maps an element ID to the element (Rank, Select, Subject, Text).

Like so you can do: 

```ts
const SubjectTree = (subject) => {
for id in subject.Order {
  const element  = subject.elements[id];
  {element.Type === SUBJECT ? <SubjectTree subject={subject}/>}
  {element.Type === RANK ? <Rank rank={element}}
  ...
}
```

Then all you need to do is to construct such Array of subjects, that contains the Order, and the Map (replaces what is done in `useConfiguration`. Also, instead of having `question.render...` it would be nicer to directly insert the component like `<Rank/>`.

web/frontend/src/pages/ballot/components/HandleAnswers.tsx
```diff
@@ -0,0 +1,121 @@
...
+    SelectAnswers: Array.from(answers.SelectAnswers),
+    RankAnswers: Array.from(answers.RankAnswers),
+    TextAnswers: Array.from(answers.TextAnswers),
+    Errors: Array.from(answers.Errors),
+  };
+
+  return newAnswers;
+}
+
+export function getIndices(
```
nkcr, line 28: Please add a comment
(same bellow)

web/frontend/src/components/utils/useConfiguration.tsx
```diff
@@ -0,0 +1,165 @@
...
+  }, [configuration]);
+
+  return {
+    sortedQuestions,
+    answers,
+    setAnswers,
+  };
+};
+
+export default useConfiguration;
```
nkcr, line 165: see my comment on `web/frontend/src/pages/ballot/Show.tsx`. It seems that this can be simplified.

web/frontend/src/pages/ballot/Show.tsx
```diff
@@ -9,32 +10,41 @@ import { ROUTE_BALLOT_INDEX } from '../../Routes';
...
-  const { postData } = usePostCall(setPostError);
+  //const [showFlash, setShowFlash] = useState(false);
   const [showModal, setShowModal] = useState(false);
   const [modalText, setModalText] = useState(t('voteSuccess') as string);
+  const { postData } = usePostCall(setPostError);
 
   useEffect(() => {
     if (postRequest !== null) {
-      setPostError('');
+      //postData(ENDPOINT_EVOTING_CAST_BALLOT, postRequest, setShowFlash);
```
nkcr, line 44: yep, we might in the end use modal instead of flash. If that's the case then commented code should be removed.

web/frontend/src/pages/election/Show.tsx
```diff
@@ -43,11 +41,11 @@ const ElectionShow: FC = () => {
     <div className="election-details-box">
       {!loading ? (
         <div>
-          <h1>{title}</h1>
+          <h1>{electionTitle}</h1>
           <div className="election-details-wrapper">
             {isResultSet ? (
               <div className="election-wrapper-child">
-                <Result resultData={result} candidates={candidates} />
+                {/* TODO: <Result resultData={result} candidates={candidates} />*/}
```
nkcr, line 48: TODOs must be linked to Github issues so we can keep track of them and also have more context (i.e. explanation). Must then be of form `TODO: #343 (add result, blabla)`.

web/frontend/src/types/configuration.ts
```diff
@@ -0,0 +1,99 @@
...
+  Title: string;
+  MaxN: number;
+  MinN: number;
+  Choices: Array<string>;
+}
+
+export interface Subject {
+  ID: ID;
+  Title: string;
+  // Order defines the order of the different question, which all have a unique
```
nkcr, line 42:
```suggestion
  // Order defines the order of the different subject elements, which all have a unique
```

web/frontend/src/types/configuration.ts
```diff
@@ -0,0 +1,99 @@
...
+  Texts?: Array<TextQuestion>;
+}
+
+// Configuration contains the configuration of a new poll.
+export interface Configuration {
+  MainTitle: string;
+  Scaffold: Array<Subject>;
+}
+
+export interface Question {
```
nkcr, line 57: This is misleading, as a `Subject` is not a question. Should better be like `SubjectElement`, or `FormElement`.

nkcr (changes requested):
Great work, this is becoming nice 👍 
Given the number of lines I couldn't make an extensive review, but you'll find already quite some elements I feel need to be updated.

web/frontend/src/pages/ballot/BallotCast.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useTranslation } from 'react-i18next';
+import { useNavigate } from 'react-router-dom';
+
+type BallotCastProps = {
+  postError: string;
+  showFlash: boolean;
+};
+
+// Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in a useEffect cleanup function.
+// Ballot@https://localhost:3000/static/js/bundle.js:5481:69
```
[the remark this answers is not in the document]
cmsigrist: Sorry, I forgot to remove this (it works, but the BallotCast might be entirely removed if we use a modal instead)

web/frontend/src/pages/ballot/Show.tsx
cmsigrist: This indeed looks a looot better ! I wasn't sure at first how to recursively display elements, so I fell across this tutorial where they flattened their data, hence the mess in useConfiguration .... 😅

web/frontend/src/components/utils/useConfiguration.tsx
```diff
@@ -0,0 +1,165 @@
...
+    ParentID: parentId,
+    Type: SUBJECT,
+    Content: subject,
+    render: renderSubject,
+  });
+
+  // The number of Question inside of a top-level Subjects (i.e a with ParentID = ROOT_ID)
+  let numberOfQuestion = 1;
+
+  if (subject.Subjects) {
```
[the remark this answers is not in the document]
cmsigrist: In the interface we created, each of the element can be undefined, so it would be necessary. But then again it might be 
better to have them as either null | Subject (text, rank, select ....) ?

web/frontend/src/types/getObjectType.ts
```diff
@@ -1,44 +1,58 @@
 import ShortUniqueId from 'short-unique-id';
-import { Rank, Select, Subject, Text } from './configuration';
+import {
```
nkcr, line 2: Something I'm not 100% convinced with, though I'm not that far:
Use instead `import * as types from './configuration';`. It makes the import list smaller, and the code easier to read since you know where the element comes from. For example:
```ts
const newSubject = (): types.Subject => {
```

what do you think ? It also applies on some other files.

nkcr (commented):
Some great improvements there, very nice !👌 
Got a couple of minor comments, after which we can merge!

web/frontend/src/pages/ballot/Show.tsx
```diff
@@ -48,108 +66,139 @@ const Ballot: FC = () => {
...
+        {element.Type === TEXT ? (
+          <Text text={element as TextQuestion} answers={answers} setAnswers={setAnswers} />
+        ) : null}
+      </div>
+    );
   };
 
-  const possibleChoice = (candidate) => {
+  const SubjectTree = (subjectElement: SubjectElement) => {
+    const subject = subjectElement as Subject;
```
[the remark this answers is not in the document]
cmsigrist: Since subject.Elements contains SubjectElements, and SubjectTree is called recursively on them I thought it made sense, but yes it's better to actually only do a cast if the SubjectElements is a Subject !

web/frontend/src/pages/ballot/components/HandleAnswers.tsx
```diff
@@ -0,0 +1,117 @@
+import { t } from 'i18next';
```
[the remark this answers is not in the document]
cmsigrist: Just to be sure files that contains Component should be  named in PascalCase and file that only contains functions or const use camelCase ?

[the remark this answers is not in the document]
nkcr: React is opinionated about that and I haven't enforced it so far. We can follow Airbnb rules that use PascalCase (https://github.com/airbnb/javascript/tree/master/react#naming).

web/frontend/src/types/getObjectType.ts
cmsigrist: Yes, good idea, some function might become a bit clunky with all the types.Text etc but I think overall it would make things clearer

nkcr: I would do it only when the import list is big, or when it adds clarity.

nkcr (approved):
LGTM ! 🚀

PR #74, opened 2022-03-30
Node exporter

jbsv:
It installs the binary and its configuration (port 9101by default). Once the package is installed, it is possible to access "http://(VMaddress):9101/metrics" from an explorer on the same network


deb-package/pkg/etc/dedis/dvoting/network.env
```diff
@@ -4,5 +4,6 @@ export uk_port="1024"
 export bridge_iface="uk0"
 
 export dela_port="9000"
-export prometheus_addr="0.0.0.0:9100"
 export proxy_addr="0.0.0.0:9080"
+export prometheus_addr="0.0.0.0:9100"
+export node_exporter_addr=":9101"
```
nkcr, line 9: for consistency, I would use `0.0.0.0:9101`

PR #77, opened 2022-04-03
Colour flash messages

cmsigrist:
Change the colour of flash messages depending on their level


nkcr (approved):
Cool, LGTM 🚀 

Some future improvements I see on the flash:

- use a different icon based on the level
- add a "close" button to discard the flash
- make the flash "overlay" the layout so that it doesn't push all the page's content

Not to be included in this PR though, let's merge it. It's good to have small feature increment.

PR #79, opened 2022-04-07
Guanyu/rewrite/scenariotest

JBGYZ:
1. Finished scenario test with http
2. added 2 fields in getelectioninfo response


integration/scenario_test.go
```diff
@@ -0,0 +1,830 @@
...
+
+		t.Parallel()
+		proxyAddr1 := "http://localhost:8081"
+		proxyAddr2 := "http://localhost:8082"
+		proxyAddr3 := "http://localhost:8083"
+
+		proxyArray := [3]string{proxyAddr1, proxyAddr2, proxyAddr3}
+
+		// ###################################### CREATE SIMPLE ELECTION ######
+		create_election_js := `{"Configuration":{"MainTitle":"electionTitle","Scaffold":[{"ID":"YWE=","Title":"subject1","Order":null,"Subjects":null,"Selects":[{"ID":"YmI=","Title":"Select your favorite snacks","MaxN":3,"MinN":0,"Choices":["snickers","mars","vodka","babibel"]}],"Ranks":[],"Texts":null},{"ID":"ZGQ=","Title":"subject2","Order":null,"Subjects":null,"Selects":null,"Ranks":null,"Texts":[{"ID":"ZWU=","Title":"dissertation","MaxN":1,"MinN":1,"MaxLength":3,"Regex":"","Choices":["write yes in your language"]}]}]},"AdminID":"adminId"}`
```
emduc, line 63: would be good to add some returns so that we can read the content of the configuration easily

emduc (commented):
I made a quick pass on it since I am starting to try and solve this shuffle problem, this new version will definitely be helpful thank you :)

I know it wasn't the case in the original scenario test but it could be nice to have the test be generic in the number of nodes  and votes, similarly to what is done in integration_test.go. It should not be too difficult to implement (just a matter of having for eg arrays of addresses instead of addr1 addr2 and addr3) but might take some time.
edit:  actually that might be more annoying since we need to launch the nodes seperately

Depending on who merges first, the problem of signed request will also arise since the proxy will soon require the requests to be signed.

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+	}
+}
+
+func testVariableNode(numNodes int, numVotes int, numElection int) func(*testing.T) {
+	return func(t *testing.T) {
+
+		proxyList := make([]string, numNodes)
+
+		for i := 0; i < numNodes; i++ {
+			if i < 10 {
```
nkcr, line 389: very hacky. What you want to do is just:
```go
 proxyList[i] = fmt.Sprintf("http://localhost:90%02d", 80+i)
```

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+			} else {
+				proxyList[i] = "http://localhost:909" + strconv.Itoa(i-10)
+			}
+			fmt.Println(proxyList[i])
+		}
+
+		var wg sync.WaitGroup
+
+		for i := 0; i < numElection; i++ {
+			fmt.Println("Starting worker", i)
```
nkcr, line 400: `t.log` instead
(same for other places)

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+	secretkeyBuf, err := hex.DecodeString("28912721dfd507e198b31602fb67824856eb5a674c021d49fdccbe52f0234409")
+	require.NoError(t, err, "failed to decode key: %v", err)
+
+	secret := suite.Scalar()
+	err = secret.UnmarshalBinary(secretkeyBuf)
+	require.NoError(t, err, "failed to Unmarshal key: %v", err)
+
+	// ###################################### CREATE SIMPLE ELECTION ######
+
+	t.Logf("Create election")
```
nkcr, line 427:
```suggestion
	t.Log("Create election")
```

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+	}
+	require.True(t, tmpCount, "front end votes are different from decrypted votes")
+
+}
+
+// -----------------------------------------------------------------------------
+// Utility functions
+func marshallBallotManual(voteStr string, pubkey kyber.Point, chunks int) (ptypes.CiphervoteJSON, error) {
+
+	var ballot = make(ptypes.CiphervoteJSON, chunks)
```
nkcr, line 703:
```suggestion
	ballots := make(ptypes.CiphervoteJSON, chunks)
```

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+		ballot[i] = ptypes.EGPairJSON{
+			K: kbuff,
+			C: cbuff,
+		}
+	}
+
+	return ballot, nil
+}
+
+func EncryptManual(message []byte, pubkey kyber.Point) (K, C kyber.Point, remainder []byte, err error) {
```
nkcr, line 743:
```suggestion
func encryptManual(message []byte, pubkey kyber.Point) (K, C kyber.Point, remainder []byte, err error) {
```

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+	return K, C, remainder, nil
+}
+
+func ChunksPerBallotManuel(BallotSize int) int {
+	if BallotSize%29 == 0 {
+		return BallotSize / 29
+	}
+
+	return BallotSize/29 + 1
+}
```
nkcr, line 767:
```suggestion
func chunksPerBallot(size int) int { return (size-1) / 29 + 1}
```

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+
+// Encode implements serde.FormatEngine
+func EncodeCiphervote(ciphervote types.Ciphervote) ([]byte, error) {
```
nkcr, line 770:
```suggestion
func encodeCiphervote(ciphervote types.Ciphervote) ([]byte, error) {
```

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+	data, err := json.Marshal(m)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to marshal cipher vote json: %v", err)
+	}
+
+	return data, nil
+}
+
+// CiphervoteJSON is the JSON representation of a ciphervote
+type CiphervoteJSON []EGPairJSON
```
nkcr, line 800: you should take the definition from `proxy/types`

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+
+// EGPairJSON is the JSON representation of an ElGamal pair
+type EGPairJSON struct {
+	K []byte
+	C []byte
+}
+
+// Unmarshal decodes the given string according to the format described in
+// "state of smart contract.md"
+func UnmarshalBallotManual(marshalledBallot string, configuration types.Configuration) (types.Ballot, error) {
```
nkcr, line 810:
```suggestion
func unmarshalBallot(ballot string, conf types.Configuration) (types.Ballot, error) {
```

nkcr: Please use the function already defined in `types/ballots.go`.

runNode.sh
```diff
@@ -0,0 +1,29 @@
...
+
+from=0
+fromb=1
+to=$1
+ne=0
+while [ $fromb -le $to ]
+do
+if [ $from -le 9 ]
+then
+gnome-terminal --tab --title="test" --command="./memcoin --config /tmp/node$fromb start --postinstall --promaddr :910$from --proxyaddr :908$from --proxykey $pk --listen tcp://0.0.0.0:200$fromb --public //localhost:200$fromb"
```
nkcr, line 17: this won't work on a standard UNIX environment (`gnome-terminal`).

integration/scenario_test.go
```diff
@@ -0,0 +1,1097 @@
...
+	}
+
+	signedJSON, err := json.Marshal(signed)
+	if err != nil {
+		return nil, xerrors.Errorf("failed to create json signed: %v", err)
+	}
+
+	return signedJSON, nil
+}
+func initDKG(secret kyber.Scalar, proxyAddr, electionIDHex string, t *testing.T) error {
```
nkcr, line 1031: missing newline above

runNode.sh
nkcr: (and it should)

nkcr: (and it should)

nkcr (approved):
LGTM 🚀

PR #81, opened 2022-04-07
Log unikernel data

jbsv:
Log data before writing to unikernel, and after receiving data from unikernel.


contracts/evoting/evoting.go
```diff
@@ -700,6 +700,9 @@ func (e evotingCommand) combineShares(snap store.Snapshot, step execution.Step)
 
 	conn.SetWriteDeadline(time.Now().Add(writeTimeout))
 
+	bytesToSend := hex.EncodeToString(buf.Bytes())
+	dela.Logger.Info().Msgf("Data to Unikernel: %v", bytesToSend)
```
nkcr, line 704: You might want to use the `.Hex()` function, something like dela.Logger.Info().Hex("out", buf.Bytes()).Msg(...).

nkcr: Same bellow.

contracts/evoting/mod_test.go
```diff
@@ -1100,6 +1102,119 @@ func TestCommand_DecryptBallots(t *testing.T) {
...
+func TestCommand_CombineSharesCanLogToDela(t *testing.T) {
+	// setup a fake logger for the test purpose
+	logBuffer := new(bytes.Buffer)
+
+	logWriter := zerolog.ConsoleWriter{
+		Out:        logBuffer,
+		TimeFormat: time.RFC3339,
+	}
+
+	dela.Logger = zerolog.New(logWriter).Level(zerolog.InfoLevel)
```
nkcr, line 1114: You don't need a  `ConsoleWriter`. You can directly use `zerolog.New(logBuffer)`. This will yield JSON, which makes it easier to compare and parse.

nkcr: When a test changes a global variable, it is always best to set it back:

```go
oldLog := dela.Logger
dela.Logger = ...
defer func() {
  dela.Logger = oldLog
}()
```

Because otherwise the next test won't have the expected logger.

contracts/evoting/evoting.go
jbsv: That works

contracts/evoting/mod_test.go
jbsv: 👍

contracts/evoting/mod_test.go
```diff
@@ -1206,13 +1207,16 @@ func TestCommand_CombineSharesCanLogToDela(t *testing.T) {
...
+	// logBuffer output example:
+	// {"level":"info","message":"using the following folder: decrypted-ballots-6475-e723-1649923012189772385/"}
+	// {"level":"info","output":"020000000a0000006465637279707465642d62616c6c6f74732d363437352d653732332d313634393932333031323138393737323338352f","message":"Data to Unikernel"}
+	// {"level":"info","input":"4f4b","message":"Data from Unikernel"}
+	logBuffers := strings.Split(logBuffer.String(), "\n")
+
+	require.Contains(t, logBuffers[1], "Data to Unikernel")
+	require.Contains(t, logBuffers[1], "output")
+	require.Contains(t, logBuffers[2], "Data from Unikernel")
+	require.Contains(t, logBuffers[2], "input")
```
nkcr, line 1219: A better way to check the content:

```go
	lines := bytes.Split(logBuffer.Bytes(), []byte("\n"))

	require.Len(t, lines, 4)

	var obj map[string]string

	err = json.Unmarshal(lines[1], &obj)
	require.NoError(t, err)

	require.Contains(t, obj["output"], hex.EncodeToString([]byte("decrypted-ballots")))
	require.Equal(t, "Data to Unikernel", obj["message"])

	err = json.Unmarshal(lines[2], &obj)
	require.NoError(t, err)

	require.Equal(t, "4f4b", obj["input"])
	require.Equal(t, "Data from Unikernel", obj["message"])
```

jbsv:
Merged main in 'unikernel-combine-shares'
then rebased 'log_unikernel_data' on new 'unikernel-combine-shares'

nkcr (approved):

PR #86, opened 2022-04-13
Fixes Frontend API requests

badrlarhdir:
Some comments for the new API request:
- [x]  Some CONST ENDPOINTS have been merged
- [x] _Not sure_: The token should still be in the header authorization w/ Bearer and we also need the userID for every API call from the frontend to the web backend : _EDIT_ : Token dropped and removed for the moment
- [x]  Status of the election ( we need that in the API doc, or ENUM/CONST somewhere to know which status corresponds to what )
- [x] Make a distinction between the hooks of getInfo and getAllInfo ( now they are the same , otherwise it breaks everything)
- [x] postData() hook function should change name


web/frontend/src/types/frontendRequestBody.ts
```diff
@@ -1,5 +1,13 @@
 import { ID } from './configuration';
 
+const INITIAL_STATUS = 0;
```
cmsigrist, line 3: Just a question but wouldn't it be easier to have an actual const enum Status ?
`const enum Status {
  Initial,
  Open,
  Closed,
  Shuffled,
  Decrypted,
  ResultAvailable,
  Canceled
}`

web/frontend/src/types/frontendRequestBody.ts
```diff
@@ -15,7 +23,14 @@ interface ElectionActionsBody {
 interface LightElectionInfo {
   ElectionID: ID;
   Title: string;
-  Status: number;
+  Status:
+    | typeof INITIAL_STATUS
```
cmsigrist, line 27: And with the enum we'd have `Status: Status`

web/frontend/src/types/frontendRequestBody.ts
badrlarhdir: Yes good point !

web/frontend/src/components/utils/Endpoints.tsx
```diff
@@ -6,14 +6,28 @@ export const ENDPOINT_USER_RIGHTS = '/api/user_rights';
...
-export const ENDPOINT_EVOTING_CREATE = '/api/evoting/create';
-export const ENDPOINT_EVOTING_RESULT = '/api/evoting/result';
-export const ENDPOINT_EVOTING_CLOSE = '/api/evoting/close';
-export const ENDPOINT_EVOTING_CANCEL = '/api/evoting/cancel';
-export const ENDPOINT_EVOTING_DECRYPT = '/api/evoting/decrypt';
-export const ENDPOINT_EVOTING_SHUFFLE = '/api/evoting/shuffle';
-export const ENDPOINT_EVOTING_CAST_BALLOT = '/api/evoting/cast';
+export const ENDPOINT_EVOTING_CREATE = '/api/evoting/elections';
+
+export const ENDPOINT_EVOTING_ELECTION = (ElectionID = undefined) =>
```
nkcr, line 11: I don't see how is it useful to have a default URL with `:ElectionID`.
To me those functions should only build the URLs epxecting a valid `ElectionID`. Additionally, we should simplify the naming and maybe use the [same as the proxy](https://github.com/dedis/d-voting/blob/adds-proxy-signed-requests/proxy/mod.go):

```ts
export const newElection = (electionID: string) => `/api/evoting/elections/${electionID}`;
export const newElectionVote = (electionID: string) => ...
```

then when we use this file we use a named import:

```ts
import * as endpoints from ...;
```

web/frontend/src/components/utils/usePostCall.tsx
```diff
@@ -1,6 +1,6 @@
 // Custom hook that post a request to an endpoint
 const usePostCall = (setError) => {
-  const postData = async (endpoint, request, setIsPosting) => {
+  const sendFetchRequest = async (endpoint, request, setIsPosting) => {
```
nkcr, line 3:
```suggestion
  return async (endpoint, request, setIsPosting) => {
```

web/frontend/src/components/utils/useChangeAction.tsx
```diff
@@ -50,22 +46,29 @@ const useChangeAction = (
     />
   );
   const [postError, setPostError] = useState(t('operationFailure') as string);
-  const { postData } = usePostCall(setPostError);
-  const simplePostRequest = {
-    method: 'POST',
+  const { sendFetchRequest } = usePostCall(setPostError);
+  const closeRequest = {
```
nkcr, line 50: You can have a generic function to update an election instead:

```ts
  const electionUpdate = async (action: string, setIsPosting) => {
    const req = {
      method: 'PUT',
      body: JSON.stringify({
        Action: action,
      }),
    };
    await sendFetchRequest(endpoints.EditElection(electionID), req, setIsPosting);
  };
```

web/frontend/src/components/utils/usePostCall.tsx
```diff
@@ -18,7 +18,7 @@ const usePostCall = (setError) => {
     }
```
nkcr, line 18: while we're at it, the `setIsPosting` should be in a `finally` block and removed elsewhere:

```ts
catch (error) {
  ...
} finally {
  setIsPosting((prev) => !prev);
}

File: web/frontend/src/types/frontendRequestBody.ts [hunk 7]
```diff
@@ -1,32 +1,42 @@
...
+  INITIAL,
+  OPEN,
+  CLOSED,
+  SHUFFLED_BALLOTS,
+  DECRYPTED_BALLOTS,
+  RESULT_AVAILABLE,
+  CANCELED,
 }
 
 interface CreateElectionBody {
```
nkcr, line 13: To match with the proxy:

```suggestion
interface NewElectionBody {
```

web/frontend/src/types/frontendRequestBody.ts
```diff
@@ -1,32 +1,42 @@
...
-  Format: any;
+  Configuration: any;
 }
 
 interface CreateElectionCastVote {
```
nkcr, line 17: NewElectionVoteBody

web/frontend/src/components/utils/usePostCall.tsx
badrlarhdir: The problem here is that we return a boolean after each setIsPosting(in either try or catch) , should I create a variable before the try and change it in the try/tach then return it in the finally ?

nkcr: Yes, you're right. I would do something like that:

```ts
  return async (endpoint, request, setIsPosting) => {
    let success = true;

    try {
      const response = await fetch(endpoint, request);

      if (!response.ok) {
        const txt = await response.text();
        throw new Error(txt);
      }

      setError(null);
    } catch (error) {
      setError(error.message);
      success = false;
    }

    setIsPosting((prev) => !prev);

    return success;
  };
```

nkcr: I pushed the update

nkcr (approved):
Good job 👍

PR #87, opened 2022-04-13
Fixes navbar UI

badrlarhdir:
- [x] Adapt the navbar for laptop
- [x] Proptypes and interface for Profile component
- [x] Adapt the navbar for mobile


web/frontend/src/layout/NavBar.tsx
```diff
@@ -55,248 +51,202 @@ const NavBar: FC = () => {
...
+                          </Popover.Button>
+                        </div>
+                      )}
+                    </div>
+                  </div>
+                </div>
+              </Popover.Panel>
+            </Transition>
+          </Popover>
+          {/* End Mobile menu */}
```
nkcr, line 189: To improve readability I would move the mobile menu to its own function component, in the same file. Same for the right/left side if this is not too complicated.

nkcr (commented):
Pretty neat 👌

nkcr (approved):
LGTM 🚀

PR #90, opened 2022-04-19
fix shuffle bug

emduc:
Fixes the bug of nodes timing out when they aren't able to get the right nonce. I ran the tests ~10-15 times and it did not reappear.


nkcr (approved):
Thank you 🙏 
This might not fix everything but it seems to already improve tests stability !

PR #78, opened 2022-04-05
Adds signature verification on proxy messages

giogio21:

PR #91, opened 2022-04-19
Implements signed requests on the proxy

nkcr:
- updates the scenario test to send signed queries
- updates the DKG proxy API to use a json type instead of plain text
- updates postinstall to include the frontend public key
- updates the scenario test to take the secret key as cli arg
- updates the proxies to accept and verify signed requests
- adds a generic function to send an http internal server error
- adds the signed request type
- updates start_test to include the proxykey

closes #78

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,11 +117,16 @@ go build ./cli/memcoin
 In three different terminal sessions, from the root folder:
 
 ```sh
-LLVL=info memcoin --config /tmp/node1 start --port 2001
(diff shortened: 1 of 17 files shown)


giogio21 (commented):
Maybe we can also update the "Run the nodes" section of README and remplace "LLVL=info memcoin --config /tmp/node1 start --port 2001" by " pk=adbacd10fdb9822c71025d6d00092b8a4abb5ebcb673d28d863f7c7c5adaddf3
LLVL=info ./memcoin --config /tmp/node1 start --postinstall --promaddr :9100 --proxyaddr :9080 --proxykey $pk --port 2001" to run a node

PR #92, opened 2022-04-20
Adds better handling of errors

nkcr:
Adds minor fixes to the frontend while working on the web-backend to support authenticated queries.

- updates the login handler to have one handler
- displays an error in case `get_personal_info` fails
- minor css fix on Modal
- fixes missing 'Content-Type' in request

### DIFF
--- a/web/frontend/src/pages/election/components/ElectionForm.tsx
+++ b/web/frontend/src/pages/election/components/ElectionForm.tsx
@@ -29,7 +29,6 @@ const ElectionForm: FC<ElectionFormProps> = ({ setShowModal, setTextModal }) => 
   // conf is the configuration object containing MainTitle and Scaffold which
   // contains an array of subject.
   const { t } = useTranslation();
-  const UserID = sessionStorage.getItem('id');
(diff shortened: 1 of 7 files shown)


web/frontend/src/pages/election/components/ElectionForm.tsx
```diff
@@ -29,19 +29,18 @@ const ElectionForm: FC<ElectionFormProps> = ({ setShowModal, setTextModal }) =>
...
 
   async function createHandler() {
     const data = {
       Configuration: marshalConfig(conf),
-      UserID,
     };
     const req = {
       method: 'POST',
       body: JSON.stringify(data),
+      headers: { 'Content-Type': 'application/json' },
```
badrlarhdir, line 43: By adding that line, we cannot longer create an election, to make it work again in the mock handler we should change the parsing of the body from `const body: NewElectionBody = JSON.parse(req.body.toString())` to 
`const body = req.body as   NewElectionBody`

badrlarhdir (approved):
Excellent, the body parsing of a new election in the mock handler should just change accordingly.

PR #93, opened 2022-04-21
Adds the "Open" action

nkcr:
- updates GET `/elections/show/:electionId` to `/elections/:electionId`
- adds missing `Content-Type` and `CORS` header
- adds better error message for `useFetchCall`

The `Open` still won't work with the proxy because we need to perform the `DK1:Init` and `DK2:Setup` steps, which are not supported yet by the frontend.


web/frontend/src/components/utils/useChangeAction.tsx
```diff
@@ -155,6 +170,17 @@ const useChangeAction = (
 
   const getAction = () => {
     switch (status) {
+      case STATUS.INITIAL:
+        return (
+          <span>
+            <button id="close-button" className="election-btn" onClick={handleOpen}>
+              {t('open')}
```
badrlarhdir, line 177: The `open` key is missing in the `en.json` file.

```ts
...
"open": "Open",
...

badrlarhdir (approved):
Excellent !

web/frontend/src/components/utils/useChangeAction.tsx
nkcr: Nice catch!

cmsigrist (approved):
LGTM !

PR #95, opened 2022-04-22
Add get results

cmsigrist:
- [x] Add a mock to show the results of an election and move the setup of a mock election in a different file for simplicity
- [x] Downloading the results now show for each question it's ID, title and the result for each candidate

- The way ballots are counted for each kind of question is liable to change


web/frontend/src/pages/election/Result.tsx
```diff
@@ -0,0 +1,168 @@
...
+import DownloadButton from 'components/buttons/DownloadButton';
+import { useTranslation } from 'react-i18next';
+import saveAs from 'file-saver';
+import { ROUTE_ELECTION_INDEX } from 'Routes';
+import { Link, useParams } from 'react-router-dom';
+import TextButton from '../../components/buttons/TextButton';
+import useElection from 'components/utils/useElection';
+import { useConfigurationOnly } from 'components/utils/useConfiguration';
+
+// Functional component that display the result of the votes
```
nkcr, line 27:
```suggestion
// Functional component that displays the result of the votes
```

web/frontend/src/pages/election/Result.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const exportData = () => {
+    const fileName = 'result.json';
+
+    // Create a blob of the data
+    const fileToSave = new Blob([JSON.stringify({ Result: result })], {
+      type: 'application/json',
+    });
+
+    saveAs(fileToSave, fileName);
+    //https://stackoverflow.com/questions/19721439/download-json-object-as-a-file-from-browser
```
nkcr, line 133:
```suggestion
    // https://stackoverflow.com/a/45594892
    saveAs(fileToSave, fileName);
```

(SO answers have a convenient "share" button that provides shorter URL ;) )

web/frontend/src/pages/election/Result.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const configuration = useConfigurationOnly(configObj);
+
+  const [rankResult, setRankResult] = useState<RankResults>(null);
+  const [selectResult, setSelectResult] = useState<SelectResults>(null);
+  const [textResult, setTextResult] = useState<TextResults>(null);
+
+  // Group the different results by the ID of the question SelectResult
+  // are mapped to 0 or 1s, such that ballots can be count more efficiently
+  const groupByID = (
+    resultMap: Map<any, any>,
```
nkcr, line 43: Looks great overall ! I have a reservation about the use of  the`any` type though. Can't we use `Map<ID, number[][]>` for the map ?
Same for `results`, it seems that it should be `string[][] | number[][] | boolean[][]`.

web/frontend/src/pages/election/Result.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const { loading, result, configObj } = useElection(electionId);
+  const configuration = useConfigurationOnly(configObj);
+
+  const [rankResult, setRankResult] = useState<RankResults>(null);
+  const [selectResult, setSelectResult] = useState<SelectResults>(null);
+  const [textResult, setTextResult] = useState<TextResults>(null);
+
+  // Group the different results by the ID of the question SelectResult
+  // are mapped to 0 or 1s, such that ballots can be count more efficiently
+  const groupByID = (
```
nkcr, line 42: Looks like `resultMap` is what the function should return:

```ts
  const groupByID = (IDs: ID[], results: any, toNumber: boolean = false): Map<ID, number[][]> => {
    const resultMap = new Map<ID, number[][]>();

    IDs.forEach((id, index) => {
      let updatedRes: number[][] = new Array<number[]>();
      let res: number[] = results[index];

      if (toNumber) {
        res = results[index].map((r: boolean) => (r ? 1 : 0));
      }

      if (resultMap.has(id)) {
        updatedRes = resultMap.get(id);
      }

      updatedRes.push(res);
      resultMap.set(id, updatedRes);
    });

    return resultMap;
  };
```

web/frontend/src/pages/election/components/SimpleTable.tsx
```diff
@@ -32,13 +32,7 @@ const SimpleTable: FC<SimpleTableProps> = ({
...
 
   const ballotsToDisplay = (elections: LightElectionInfo[]) => {
-    let dataToDisplay = [];
-    elections.forEach((election) => {
-      if (election.Status === statusToKeep) {
-        dataToDisplay.push(election);
-      }
-    });
-    return dataToDisplay;
+    return elections.filter((election) => election.Status === statusToKeep);
```
nkcr, line 35: 👍

nkcr (commented):
The result page looks great :)

web/frontend/src/pages/election/Result.tsx
cmsigrist: Haha true, admittedly I only copy pasted the code and the link that was previously in DownloadResult.tsx 😁

web/frontend/src/pages/election/Result.tsx
cmsigrist: The answer is pretty simple, I forgot we could type objects with ORs ! 😔

web/frontend/src/pages/election/Result.tsx
cmsigrist: I'm not sure it can be done since groupByID is called in a loop in groupResultsByID and groupResultsByID returns the maps

web/frontend/src/pages/election/components/SelectResult.tsx
```diff
@@ -0,0 +1,39 @@
...
+          <ProgressBar isBest={isBest}>{percent}</ProgressBar>
+        </React.Fragment>
+      );
+    });
+  };
+
+  return (
+    <div
+      className="grid [grid-template-columns:_min-content_auto] gap-1 items-center"
+      key={select.ID}>
```
nkcr, line 33: Doesn't seem necessary.

nkcr (approved):
LGTM 🚀

PR #96, opened 2022-04-26
Change flash icons, position and add close button

cmsigrist:
- [x] Change the icons depending on the FlashLevel of the message
- [x] Add a visibility property to FlashMessage such that clicking on close sets it to false (since deleting it from the array does not work with the Timeout) 
- [x] The flash message now appears on top of the navbar rather than pushing everything down


nkcr (approved):
Thank you, this is much better now 🙏 
I added a small animation.

cmsigrist:
> @cmsigrist I updated a bit the code. Please have a look and merge : )

This is so much better ! 
Also we discussed the position of the flash message with Badr and we thought that it was maybe a bit less intrusive on the right of the screen. I added another animation :D

nkcr:
I'm not a big fan of this positioning. Flash messages, as opposed to popup/modal, should not overlay the main content of the website. Making users free to "just ignore" them. On small screens this is not the case. We need to close the flash in order to use the page:

Screenshot 2022-04-27 at 09 23 29
Screenshot 2022-04-27 at 09 32 07

Also, on very big screen the user needs to go all the way to the right in order to close the flash message. Apart from making the flash less visible (here we see it well because the page's content is mostly empty), this is not convenient:

Screenshot 2022-04-27 at 09 39 48

nkcr:
👌

PR #97, opened 2022-04-26
Displays correct actions for each role

badrlarhdir:
Displays the correction action for each role.
Note that the AuthContext is not well implemented and we should have access to the setAuth when calling  `useContext(AuthContext)`, to better handle the state of the Auth when **_Logged In / Logged Out_**, another PR should change that (and also pass the roles to an enum to remove hardcoding the values inside the code).


web/frontend/src/layout/NavBar.tsx
```diff
@@ -228,8 +228,12 @@ const NavBar: FC = () => {
     } else {
       fctx.addMessage(t('logOutSuccessful'), FlashLevel.Info);
     }
-
+    // TODO: should be a setAuth function passed to AuthContext rather than
```
nkcr, line 231: Indeed we should add a `logout()` function to `AuthState` that does that.

web/frontend/src/components/utils/useChangeAction.tsx
```diff
@@ -16,6 +17,7 @@ const useChangeAction = (
   setShowModalError: (willShow: boolean) => void
 ) => {
   const { t } = useTranslation();
+  const { role } = useContext(AuthContext);
```
nkcr, line 20: In the rest of the code base it is called `authCtx`. Let's stick to that for consistency.

web/frontend/src/components/utils/useChangeAction.tsx
```diff
@@ -169,69 +172,99 @@ const useChangeAction = (
   };
 
   const getAction = () => {
```
nkcr, line 174: This is good enough for now as it is legacy code. But instead of returning a single action based on the role and status, each action should be rendered based on those two parameters. Like so we would have each component defined separately, and we could have not only one action. For example:

```ts
const OpenBTN = ({status}) => {
  const { authCtx } = useContext(AuthContext);

  const isAuthorized = () : boolean {
    return authCtx.role === 'admin' || authCtx.role === 'operator'
  }

  return (
    {isAuthorized && status === STATUS.INITIAL ? (
      <button>...</button>
    ) : null}
  )
}

const CloseBTN = ({status}) => {
  const { authCtx } = useContext(AuthContext);

  const isAuthorized = () : boolean {
    return authCtx.role === 'admin' || authCtx.role === 'operator'
  }

  return (
    {isAuthorized && status === STATUS.OPEN ? (
      <button>...</button>
    ) : null}
  )
}
```

And then, in the main page you have:

```ts
  <div>
    ...
    <OpenBTN status={status} />
    <CloseBTN status={status} />
    ...
  </div>
```

web/frontend/src/components/utils/CancelButton.tsx
```diff
@@ -0,0 +1,14 @@
+import { AuthContext } from 'index';
+import { useContext } from 'react';
+import { STATUS } from 'types/electionInfo';
+
+const CancelButton = ({ status, handleCancel, t }) => {
```
nkcr, line 5: As for `authCtx`, `t` is not needed as argument:

```
const { t } = useTranslation();
```

nkcr (approved):
Good ! Minor comment, otherwise LGTM 🚀

PR #98, opened 2022-04-27
Added status metrics to d-voting

jbsv:
First step is to add the d-voting status and test it in integration test.
More tests to come ASAP.

### DIFF
--- a/contracts/evoting/evoting.go
+++ b/contracts/evoting/evoting.go
@@ -240,7 +240,7 @@ func (e evotingCommand) castVote(snap store.Snapshot, step execution.Step) error
 		return xerrors.Errorf("failed to set value: %v", err)
 	}
 
-	PromElectionBallots.WithLabelValues(election.ElectionID).Inc()
+	PromElectionBallots.WithLabelValues(election.ElectionID).Set(float64(len(election.Suffragia.Ciphervotes)))
 
 	return nil
 }
--- a/contracts/evoting/evoting.go
+++ b/contracts/evoting/evoting.go
@@ -390,7 +390,7 @@ func (e evotingCommand) shuffleBallots(snap store.Snapshot, step execution.Step)
 
 	election.ShuffleInstances = append(election.ShuffleInstances, currentShuffleInstance)
 
-	PromElectionShufflingInstances.WithLabelValues(election.ElectionID).Add(float64(len(election.ShuffleInstances)))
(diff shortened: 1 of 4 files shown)


contracts/evoting/evoting.go
```diff
@@ -101,6 +101,8 @@ func (e evotingCommand) createElection(snap store.Snapshot, step execution.Step)
 		ShuffleThreshold: threshold.ByzantineThreshold(roster.Len()),
 	}
 
+	PromElectionStatus.Set(float64(election.Status))
```
pierluca, line 104: To manage multiple concurrent elections, using a label would be necessary.
```suggestion
	PromElectionStatus.WithLabelValues(election.ElectionID).Set(float64(election.Status))
```

(here and elsewhere)

contracts/evoting/mod.go
```diff
@@ -19,6 +21,33 @@ import (
 	_ "github.com/dedis/d-voting/contracts/evoting/json"
 )
 
+var (
+	PromElectionStatus = prometheus.NewGauge(prometheus.GaugeOpts{
+		Name: "dvoting_evoting_status",
+		Help: "status of election",
+	})
```
pierluca, line 28: likewise, here and elsewhere, it would have to look something like
```suggestion
	PromElectionStatus = prometheus.NewGauge(prometheus.GaugeOpts{
		Name: "dvoting_evoting_status",
		Help: "status of election",
	},
	[]string{"election"})
```

contracts/evoting/mod.go
```diff
@@ -19,6 +21,33 @@ import (
...
+
+	PromElectionBallots = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "dvoting_evoting_ballots",
+		Help: "number of cast ballots",
+	})
+
+	PromElectionShufflingInstances = prometheus.NewCounter(prometheus.CounterOpts{
+		Name: "dvoting_evoting_shufflings",
+		Help: "number of shuffling instances",
+	})
```
pierluca, line 38: counter entails that this always goes up (for a given election, using labels).
Is it the case? could there be an operational change in an election which reduces this number? @nkcr @jbsv

pierluca (commented):
Labels seem to be the main thing missing, and I have a doubt about shuffling instances, but the rest LGTM !

contracts/evoting/mod.go
jbsv: Unless we have a reshuffle, I don't see it going down.

nkcr (approved):

PR #101, opened 2022-04-27
Mock admin

cmsigrist:
Add a mock for the Admin page

### DIFF
--- a/web/frontend/src/mocks/handlers.ts
+++ b/web/frontend/src/mocks/handlers.ts
@@ -13,16 +13,21 @@ import {
   EditElectionBody,
   NewElectionBody,
   NewElectionVoteBody,
+  NewUserRole,
+  RemoveUserRole,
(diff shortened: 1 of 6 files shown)


web/frontend/src/mocks/handlers.ts
```diff
@@ -157,4 +162,22 @@ export const handlers = [
 
     return res(ctx.status(200), ctx.text('Action successfully done'));
   }),
+
+  rest.get(endpoints.ENDPOINT_USER_RIGHTS, (req, res, ctx) => {
```
nkcr, line 166: Makes me realize we should also define the admin API in the doc.

nkcr: Oh, this is indeed in `docs/frontend_doc.md` 🙈

nkcr (approved):

PR #102, opened 2022-04-27
Fix the ballot encoding

cmsigrist:
Fix the way the encoded ballot is divided into chunks of 29 bytes (counting the characters resulted in a bug where some chunks
where longer than 29 bytes because of special characters).
A workaround is to first encode into bytes the ballot, divide the resulting array of bytes into chunks of 29 bytes and then decode each chunks back into a string (of 29 bytes) ...


nkcr:
Made an update, when can keep the chunks as `Uint8Array[]` all the way until encryption.

nkcr (approved):

cmsigrist:
> Made an update, we can keep the chunks as `Uint8Array[]` all the way until encryption.

Excellent (I wasn't sure if it would work with bytes hence the decoding back into strings) !

PR #105, opened 2022-04-28
Fix the status enum

cmsigrist:
Add intermediate status (i.e being setup, being decrypted, being shuffled) as well as a Setup status. These intermediate states correspond to times when polling or SSE will have to be implemented to know when to go to the next status (as such their numbering from 7 to 10 does not have an equivalent in the backend).
Actions will be modified accordingly in another PR.


web/frontend/src/components/utils/CloseButton.tsx
```diff
@@ -2,15 +2,16 @@ import { AuthContext } from 'index';
...
 import { useTranslation } from 'react-i18next';
 import { STATUS } from 'types/electionInfo';
+import { Role } from 'types/userRole';
 
 const CloseButton = ({ status, handleClose }) => {
   const authCtx = useContext(AuthContext);
   const { t } = useTranslation();
 
-  const isAuthorized = authCtx.role === 'admin' || authCtx.role === 'operator';
+  const isAuthorized = authCtx.role === Role.Admin || authCtx.role === Role.Operator;
```
badrlarhdir, line 11: For consistency with STATUS and the rest of the CONSTS/ENUMS shouldn't Role be in upper case `ROLE` ?

badrlarhdir (approved):
Looks good to me.

web/frontend/src/components/utils/CloseButton.tsx
cmsigrist: Right !

web/frontend/src/types/election.ts
```diff
@@ -0,0 +1,76 @@
+import { ID } from './configuration';
+
+export const enum STATUS {
+  // Initial is when the election has just been created
+  Initial = 0,
+  // InitializedNode is when the nodes has been initialized by the dkg service
+  InitializedNodes = 7,
```
nkcr, line 7: if this is an internal status for the frontend this is fine. Otherwise the DKG statuses are defined there: https://github.com/dedis/d-voting/blob/065815b355841e48429036e311348e1542263087/services/dkg/mod.go#L20

nkcr (approved):
🚀

web/frontend/src/types/election.ts
cmsigrist: They're are internal status for the frontend but it would probably be better to rename them so that they match with the DKG statuses

PR #109, opened 2022-05-02
adds polling on the nodes' state instead of sleeping

emduc:
Replaces the different sleeps in the integration test by busy polling on the nodes' states; we wait that all the nodes reached a concensus on the desired state, within a given timeout. Timing out raises an error but does not stop the test, this is to get a better understanding of the error: eventually the test will stop on its own but the additional logs after the timeout are useful to understand what goes wrong.

### DIFF
--- a/integration/integration_test.go
+++ b/integration/integration_test.go
@@ -111,7 +111,7 @@ func getIntegrationTest(numNodes, numVotes int) func(*testing.T) {
 		err = closeElection(m, electionID, adminID)
 		require.NoError(t, err)
 
-		time.Sleep(time.Second * 1)
+		waitForStatus(types.Closed, electionFac, electionID, nodes, t, numNodes, 2*time.Second)
 
 		// ##### SHUFFLE BALLOTS #####
 		t.Logf("initializing shuffle")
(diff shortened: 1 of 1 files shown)


integration/integration_test.go
```diff
@@ -570,3 +573,32 @@ func closeNodes(t *testing.T, nodes []dVotingCosiDela) {
 func encodeID(ID string) types.ID {
 	return types.ID(base64.StdEncoding.EncodeToString([]byte(ID)))
 }
+
+// waitForStatus polls the node until they all updated to the expected status
```
nkcr, line 577:
```suggestion
// waitForStatus polls the nodes until they all updated to the expected status
```

integration/integration_test.go
```diff
@@ -570,3 +573,32 @@ func closeNodes(t *testing.T, nodes []dVotingCosiDela) {
...
+// for the given election. An error is raised if the timeout expires.
+func waitForStatus(status types.Status, electionFac types.ElectionFactory, electionID []byte,
```
nkcr, line 579: As a common convention, `testing.T` should be the first argument.

integration/integration_test.go
```diff
@@ -570,3 +573,32 @@ func closeNodes(t *testing.T, nodes []dVotingCosiDela) {
...
+	nodes []dVotingCosiDela, t *testing.T, numNodes int, timeOut time.Duration) {
+	// setup a timer to fail the test in case we never reach the status
+	timer := time.NewTimer(timeOut)
+	go func() {
+		<-timer.C
+		t.Errorf("timed out while waiting for status %d", status)
+	}()
+
+	loop := true
```
nkcr, line 588: I'm suggesting doing as follow to simulate the "do while":
```go
	isOK := func() bool {
		for _, node := range nodes {
			election, err := getElection(electionFac, electionID, node.GetOrdering())
			require.NoError(t, err)

			if election.Status != status {
				return false
			}
		}

		return true
	}

	for !isOK() {
		time.Sleep(time.Millisecond * 100)
	}
```

nkcr (commented):
Thanks 💯

nkcr (approved):

PR #114, opened 2022-05-04
Add frontend Actions

cmsigrist:
- [x] Add new actions to initialize the nodes participating in an election and to set them up
- [x] Remove the actions from the election table, and replace them with QuickActions
- [x] Add polling on the status of the nodes and the status of an election to know when it has been successfully completed
- [x] Add a local storage for OngoingActions and the node to setup such that if the user reloads or refresh the ElectionShow page, the actions does not need to be re-triggered and the polling restarts.
- [x] Add a filter by Status on the ElectionTable 
- [x] Display the nodes participating in the election and their status in ElectionShow 
- [x] Add a DKG table to display the status of the nodes and their associated proxy, the proxy address can be edited in the table, this will be added in another PR in the Admin page 

- For the moment, the Status shown in the ElectionTable are the complete election "flow" (not the lighter version as in ElectionShow when a user is not an Admin or Operator)


nkcr (commented):
Did a first pass on it, this is quite massive! 😯
It feels there could be some significant simplifications in the code (to be discussed in live), but great job overall 👍

web/frontend/src/components/utils/useChangeAction.tsx
```diff
@@ -63,189 +105,470 @@ const useChangeAction = (
...
+        break;
+      case OngoingAction.Combining:
+        pollElectionStatus(Status.PubSharesSubmitted, Status.ResultAvailable);
+        setResultAvailable(true);
+        break;
+      default:
+        break;
+    }
+
+    return () => {
```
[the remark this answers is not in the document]
cmsigrist: It's (apparently) called a useEffect cleanup function, and it's called right before the component unmounts, which is when we want to stop polling or fetching
https://blog.logrocket.com/understanding-react-useeffect-cleanup-function/

web/frontend/src/pages/election/Index.tsx
```diff
@@ -17,30 +29,151 @@ const ElectionIndex: FC = () => {
...
+        }
+      };
+
+      const promises = elections
+        .filter((election) => election.Status === Status.Initial)
+        .map((election) => {
+          return fetchProxies(election);
+        });
+
+      Promise.all(promises)
```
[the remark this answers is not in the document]
cmsigrist: Yes this is indeed problematic. I think I tried to synchronize them because there was some weird things happening with the states (the status was not re-rendering on each update, etc...). I have just refactored the code so there's no synchronization anymore. But I don't know when to set the DKGLoading to False (e.g all the elections that needs to have the status of their nodes have them). So the status sometimes changes after the page has loaded...

web/frontend/src/pages/admin/DKGTable.tsx
```diff
@@ -0,0 +1,212 @@
...
+            </th>
+            <th scope="col" className="px-6 py-3">
+              <span className="sr-only">Edit</span>
+            </th>
+          </tr>
+        </thead>
+        <tbody>
+          <>
+            {proxyAddresses !== null && DKGStatuses !== null
+              ? Array.from(proxyAddresses).map(([node, _proxy], index) => (
```
nkcr, line 148: The map should look as follow:

```ts
Array.from(proxyAddresses).map(([node, _proxy], index) => (
  <ProxyRow node={node} />
)}
```

nkcr: It should simplify much the code and remove the need to have those `Map`.

nkcr (commented):
Very nice improvement ! 💯

web/frontend/src/components/utils/Endpoints.tsx
```diff
@@ -10,9 +10,19 @@ export const newElection = '/api/evoting/elections';
...
 export const editDKGActors = (ElectionID: string) =>
   `/api/evoting/services/dkg/actors/${ElectionID}`;
+// initialize the nodes
+export const dkgActors = `/api/evoting/services/dkg/actors`;
+
+export const getDKGActors = (Proxy: string, ElectionID: string) =>
+  encodeURI(`${Proxy}/evoting/services/dkg/actors/${ElectionID}`);
+
+// get the proxies' addresses from the web backend
+export const getProxiesAddresses = (ElectionID: string) => `/api/elections/${ElectionID}/proxies`;
```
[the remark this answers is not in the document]
cmsigrist: Is the response to getProxiesAddresses still this ?
```
{"Proxies": [
  {"123.456.78.9:9000": "https://example.com/"},
  {"111.222.333.444:9000": "https://blabla.ch/"}
]}
```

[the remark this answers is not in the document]
nkcr: Yes, this would be the result of a 

```
GET
/api/proxies
```

web/frontend/src/pages/admin/DKGTable.tsx
```diff
@@ -0,0 +1,81 @@
+import React, { FC, useContext, useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import ProxyRow from './ProxyRow';
+import * as endpoints from 'components/utils/Endpoints';
+import useFetchCall from 'components/utils/useFetchCall';
+import { FlashContext, FlashLevel } from 'index';
+import Loading from 'pages/Loading';
+
+const DKGTable: FC = () => {
```
nkcr, line 9: Should be added on the admin page.
It is missing a button to add and delete a mapping.

web/frontend/src/pages/election/components/ElectionRow.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { LightElectionInfo } from 'types/election';
+import { Link } from 'react-router-dom';
+import ElectionStatus from './ElectionStatus';
+import QuickAction from './QuickAction';
+
+type ElectionRowProps = {
+  election: LightElectionInfo;
+};
+
+const ElectionRow: FC<ElectionRowProps> = ({ election }) => {
```
nkcr, line 11: Yes ! That's the kind of component pleasant to review 😄

nkcr (approved):
What a massive work ! I left two comments that can be addressed in a separate PR. We should not wait any longer to merge this work :)

web/frontend/src/pages/admin/DKGTable.tsx
cmsigrist: Right I'm doing it right now 😁

PR #118, opened 2022-05-06
Adds the functionality to delete an election

nkcr:
Adds the functionality to delete an election. Even if everything is on the blockchain, deleting an election will remove it from the global state.

Updates all components (smart contract, proxy, web-backend, web-frontend, API doc) accordingly.

### DIFF
--- a/proxy/election.go
+++ b/proxy/election.go
@@ -22,6 +22,7 @@ import (
 	"go.dedis.ch/dela/core/txn/pool"
 	"go.dedis.ch/dela/serde"
 	"go.dedis.ch/kyber/v3"
+	"go.dedis.ch/kyber/v3/sign/schnorr"
 	"golang.org/x/xerrors"
 )
 
@@ -147,7 +148,7 @@ func (h *election) NewElectionVote(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !elecMD.ElectionsIDs.Contains(electionID) {
+	if elecMD.ElectionsIDs.Contains(electionID) < 0 {
 		http.Error(w, "the election does not exist", http.StatusNotFound)
 		return
 	}
@@ -214,7 +215,7 @@ func (h *election) EditElection(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if !elecMD.ElectionsIDs.Contains(electionID) {
+	if elecMD.ElectionsIDs.Contains(electionID) < 0 {
 		http.Error(w, "the election does not exist", http.StatusNotFound)
 		return
 	}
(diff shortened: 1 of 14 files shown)


web/frontend/src/components/utils/DeleteButton.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const authCtx = useContext(AuthContext);
+  const { t } = useTranslation();
+
+  const isAuthorized = authCtx.role === ROLE.Admin || authCtx.role === ROLE.Operator;
+
+  return isAuthorized ? (
+    <button onClick={handleDelete}>
+      <div className="whitespace-nowrap inline-flex items-center justify-center px-4 py-1 mr-2 border border-gray-300 text-sm rounded-full font-medium text-gray-700">
+        <TrashIcon className="-ml-1 mr-2 h-5 w-5 text-gray-700" aria-hidden="true" />
+        {t('delete')}
```
badrlarhdir, line 17: Combine is still missing in the language json (sorry I forgot to add in the previous PR)

web/frontend/src/pages/election/components/Action.tsx
```diff
@@ -28,11 +33,28 @@ const Action: FC<ActionProps> = ({ status, electionID, setStatus, setResultAvail
     setShowModalError
   );
 
+  const deleteElection = async () => {
```
badrlarhdir, line 36: There should be a function in the handler in the mock API to "make it work".

web/backend/src/Server.ts
```diff
@@ -287,6 +287,60 @@ app.use('/api/evoting/*', (req, res, next) => {
   }
 });
 
+// https://stackoverflow.com/a/1349426
+function makeid(length: number) {
```
badrlarhdir, line 291: Since we are already using `short-unique-id` on the frontend, maybe using it too on the web backend should be better.

badrlarhdir (changes requested):
The web part looks good to me, maybe the implementation in the mock handler would make everything work on the frontend, at least with the mock API.

web/backend/src/Server.ts
nkcr: In fact this code is a hack for a demo that I will present next week. It will be removed afterward.

giogio21 (commented):
The backend part looks good to me

proxy/election.go
```diff
@@ -453,6 +454,61 @@ func (h *election) Elections(w http.ResponseWriter, r *http.Request) {
...
+		InternalError(w, r, xerrors.Errorf("failed to marshal DeleteElection: %v", err), nil)
+		return
+	}
+
+	_, err = h.submitAndWaitForTxn(r.Context(), evoting.CmdDelteElection, evoting.ElectionArg, data)
+	if err != nil {
+		http.Error(w, "failed to submit txn: "+err.Error(), http.StatusInternalServerError)
+		return
+	}
+}
```
emduc, line 510: I'm not entirely familiar with the new format of the proxy yet, we don't need to write back here?

emduc (approved):
Nice and sweet 💯

web/backend/src/Server.ts
nkcr: (but this is a very valid comment otherwise!)

proxy/election.go
nkcr: No, there is no data to return. Of course if something went wrong the proxy will return an HTTP error (error 500).

PR #120, opened 2022-05-09
Adds redirect after actions

### LINKED ISSUE #117: Frontend correct redirection after Action
Fixes to have correct redirection
- When creating an election the user should be redirected to the election show page of this new election
- When casting a vote the user should be redirected to the correct page ( show page or elections table depending on where he came from )
- The back button of show results should redirect the user to the correct page  ( show page or elections table depending on where he came from )

badrlarhdir:
- [x]  When creating election returns to the election show page *
- [x]  When going back from the results should return to the correct `from` page
- [x]  When casting a vote should return to the correct page ( show page or elections table ) 

Fixed : when redirecting the modal doesn't show itself since it redirects to the correct page which doesn't know if it was a redirect or not. I think the redirect should occur after closing the modal, but we will need to adapt the modal to make it work.

The bonus will be done in a separate PR with the roles auth context.

close #117


web/frontend/src/pages/election/components/ElectionForm.tsx
```diff
@@ -65,9 +67,11 @@ const ElectionForm: FC<ElectionFormProps> = ({ setShowModal, setTextModal }) =>
         setShowModal(true);
       } else {
         const response = await res.json();
-        setTextModal(`Success creating an election ! ElectionID : ${response.ElectionID}`);
-        setShowModal(true);
+        // TODO: remove the following lines if we don't need a confirmation modal anymore
+        // setTextModal(`Success creating an election ! ElectionID : ${response.ElectionID}`);
+        // setShowModal(true);
```
nkcr, line 72: I would expect to be redirected to the election page once I confirm the modal.

nkcr (approved):

PR #127, opened 2022-05-17
Fix Front UI create election

badrlarhdir:
- [x]  Main page
  - [x]  remove drag and drop for importing JSON and replace it with only text
  - [x]  padding top main title
  - [x]  Block default input of titles so i can remove no main title etc.
  - [x]  default input when importing json is already validated
  - [x]  Line for *remove subject* in dropdown
  - [x] Rank min/max values to be only one
  - [x] Tab for preview in phones
  - [x] Problem : working on both phone and laptop without having two different states
- [x]  Modal
  - [x]  2 col modal
  - [x]  Add type Icons in modal create question
  - [x]  MaxN & minN more relevant
  - [x]  Validation on each question before saving
  - [x]  Add confirm modal for remove subject
  - [ ]  Preview of the question inside the modal ( TODO later )


web/frontend/src/pages/election/components/Question.tsx
```diff
@@ -1,168 +1,69 @@
...
+              <DisplayTypeIcon Type={Type} />
             </div>
-          </Disclosure.Panel>
+            <div className="pt-1.5 max-w-md pr-8 truncate">
+              {Title.length ? Title : `Enter ${Type} title`}
+            </div>
+          </div>
+
+          <div className="flex mt-2 ml-2">
+            <div className="ml-8">{Type.charAt(0).toUpperCase() + Type.slice(1)}</div>
```
nkcr, line 57: I still think this is not necessary to display the question type in plain text, as there is already the icon.

web/frontend/src/pages/election/components/SubjectDropdown.tsx
```diff
@@ -0,0 +1,73 @@
+import { Menu, Transition } from '@headlessui/react';
+import { DotsVerticalIcon } from '@heroicons/react/outline';
+import { Fragment } from 'react';
+import { useTranslation } from 'react-i18next';
+
+const removeElementNames = ['removeSubject', 'removeselect', 'removerank', 'removetext'];
+
+const SubjectDropdown = ({ dropdownContent }) => {
```
nkcr, line 8: the prop type is missing

nkcr (commented):
Good job 👍 
The input fields would render better with a bit of padding, I think :)

web/frontend/src/pages/election/components/Question.tsx
badrlarhdir: I think we made the redundancy on purpose in case someone doesn't understand the icon @cmsigrist .

nkcr (approved):

PR #128, opened 2022-05-17
Fix the UI of the result and vote page

cmsigrist:
Some quick fixes of the UI for the result and vote pages, so that the UI is more coherent.
Refactor the BallotShow so that it can also be used as a preview during the ElectionCreate


nkcr (approved):
Nice ! 💯 
Can be merged once the conflicts are solved.

PR #130, opened 2022-05-30
Adds proxies handling on the web backend

nkcr:


docs/frontend_doc.md
```diff
@@ -264,6 +264,10 @@ Return:
 
 ```
 
+`NodeAddr` MUST be URL encoded. For example `123.456.9:9000` MUST become
```
cmsigrist, line 267: I used `encodeURI()` to encode the entire endpoint, but it doesn't escape the ':' character ! Should I modify the endpoints and push the change on this branch ?

cmsigrist (approved):
LGTM !

PR #134, opened 2022-06-02
Updates web packages 

nkcr:


web/frontend/package.json
```diff
@@ -14,12 +14,10 @@
   },
   "dependencies": {
     "@dedis/kyber": "^3.4.4",
-    "@emotion/react": "^11.7.1",
-    "@emotion/styled": "^11.6.0",
     "@headlessui/react": "^1.5.0",
     "@heroicons/react": "^1.0.6",
-    "@material-ui/core": "^4.12.3",
-    "@mui/material": "^5.4.0",
+    "@material-ui/core": "^4.12.4",
```
badrlarhdir, line 19: I don't see any imports inside the code of material UI , I think we can remove it .

badrlarhdir (approved):
Looks good to me !

web/frontend/package.json
nkcr: Great, nice catch !

web/backend/src/Server.ts
```diff
@@ -7,8 +7,19 @@ import kyber from '@dedis/kyber';
...
-const config = require('../config.json');
+const MemoryStore = createMemoryStore(session);
+
+// store is used to store the session
+const store = new MemoryStore({
+  checkPeriod: 86400000, // prune expired entries every 24h
+});
+
+// Keeps an in-memory mapping between a SCIPER (userid) and its opened session
+// IDs. Needed to invalidate the sessions of a used when its role changes. The
```
cmsigrist, line 20:
```suggestion
// IDs. Needed to invalidate the sessions of a user when its role changes. The
```

web/backend/src/Server.ts
```diff
@@ -7,8 +7,19 @@ import kyber from '@dedis/kyber';
...
+// value is a bag of sessions IDs.
+const sess2sciper = new Map<number, Map<string, undefined>>();
```
cmsigrist, line 22: What is the undefined field of the inner map ?

cmsigrist (approved):

web/backend/src/Server.ts
nkcr: The map is used as a bag/set. Javascript doesn't have a native data structure for that, so I am using a map where only the keys are used.

cmsigrist: Oh ok I see, although sets do exist in Javascript which should work for the session IDs, at least if it doesn't matter if duplicate session IDs for a user only appear once in the set ?
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

nkcr: Thanks for pointing that, I overlooked this native `Set` data-structure, which exactly fits the purpose !

PR #136, opened 2022-06-03
Gets the default proxy with the backend

nkcr:
By default the app uses the default proxy provided by the web backend. However users can freely change it. The proxy is saved on the session store to avoid fetching the proxy each time.


web/frontend/src/components/utils/proxy.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const [proxy, setProxy] = useState<string>(pctx.getProxy());
+  const [inputVal, setInputVal] = useState('');
+  const [inputChanging, setInputChanging] = useState(false);
+  const [inputWidth, setInputWidth] = useState(0);
+
+  const proxyTextRef = createRef<HTMLDivElement>();
+
+  const fetchFromBackend = async () => {
+    try {
+      const response = await fetch('/api/config/proxy');
```
cmsigrist, line 20:
```suggestion
      const response = await fetch(endpoints.getProxyConfig);
```

web/frontend/src/components/utils/proxy.tsx
```diff
@@ -0,0 +1,103 @@
+import { CheckIcon, PencilIcon, RefreshIcon } from '@heroicons/react/outline';
+import { FlashContext, FlashLevel, ProxyContext } from 'index';
+import { ChangeEvent, FC, createRef, useContext, useEffect, useState } from 'react';
+
```
cmsigrist, line 4:
```suggestion
 import * as endpoints from './Endpoints';
 
```

web/frontend/src/index.tsx
```diff
@@ -159,6 +181,22 @@ const AppContainer = () => {
     },
   };
 
+  const setDefaultProxy = async () => {
+    let proxy = sessionStorage.getItem('proxy');
+
+    if (proxy === null) {
+      const response = await fetch('/api/config/proxy');
```
cmsigrist, line 188: endpoints.getProxyConfig can alos be called here :D

cmsigrist (approved):
LGTM ! :)

PR #144, opened 2022-06-20
Fix the ballot encoding

cmsigrist:
Fix the ballot encoding:  padding is added until the size of the encoded ballot is equal to the `BallotSize`.


emduc (approved):
Seems all good to me ! Thanks for the update

nkcr (approved):

PR #152, opened 2022-06-24
Display the number of ballot cast so far on election page

cmsigrist:
- [x] Display the number of ballot that were cast for an election
- [x] Display the ID of the users who voted for the election
- [x] Display the status of the nodes in a table
- [x] Make each row independent, they now each fetch their status
- [x] Add a timeout when the web backend or the proxy does not respond after 10 seconds


nkcr (approved):
Nice 💯

PR #153, opened 2022-06-26
Fix web backend to redirect DKG setup requests

cmsigrist:
I am absolutely not sure if this is correctly written !
But currently DKG setup requests are not redirected, they use the default proxy, the idea is to redirect them to the correct proxy (the one provided in the input json). 
As the `regex` used to check if it's an initialisation request does not match the setup i.e `const regex = /\/evoting\/services\/dkg\/actors$/` does not match with the setup endpoint: `/api/evoting/services/dkg/actors/${ElectionID}`

### DIFF
--- a/web/backend/src/Server.ts
+++ b/web/backend/src/Server.ts
@@ -408,7 +408,7 @@ function sendToDela(dataStr: string, req: express.Request, res: express.Response
   }
 
   // in case this is a DKG setup request, we must update the payload.
-  const dkgSetupRegex = /\/evoting\/services\/dkg\/actors\//;
+  const dkgSetupRegex = /\/evoting\/services\/dkg\/actors\/.*$/;
   if (uri.match(dkgSetupRegex)) {
     const dataStr2 = JSON.stringify({ Action: req.body.Action });
     payload = getPayload(dataStr2);
--- a/web/backend/src/Server.ts
+++ b/web/backend/src/Server.ts
@@ -400,13 +400,6 @@ function sendToDela(dataStr: string, req: express.Request, res: express.Response
   // we strip the `/api` part: /api/election/xxx => /election/xxx
   let uri = process.env.DELA_NODE_URL + req.baseUrl.slice(4);
 
-  // in case this is a DKG setup request, we must update the payload.
-  const regex = /\/evoting\/services\/dkg\/actors/;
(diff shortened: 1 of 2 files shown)


web/backend/src/Server.ts
```diff
@@ -400,14 +400,24 @@ function sendToDela(dataStr: string, req: express.Request, res: express.Response
   // we strip the `/api` part: /api/election/xxx => /election/xxx
   let uri = process.env.DELA_NODE_URL + req.baseUrl.slice(4);
 
-  // in case this is a DKG  init request, we must extract the proxy addr and
-  // update the payload.
-  const regex = /\/evoting\/services\/dkg\/actors$/;
+  // in case this is a DKG setup request, we must update the payload.
+  const regex = /\/evoting\/services\/dkg\/actors/;
```
nkcr, line 404:
```suggestion
  const regex = /\/evoting\/services\/dkg\/actors$/;
```

web/backend/src/Server.ts
```diff
@@ -400,14 +400,24 @@ function sendToDela(dataStr: string, req: express.Request, res: express.Response
...
+
+  // in case this is a DKG init request, we must also update the payload.
+  const dkgInitRegex = /\/evoting\/services\/dkg\/actors$/;
+  if (uri.match(dkgInitRegex)) {
     const dataStr2 = JSON.stringify({ ElectionID: req.body.ElectionID });
     payload = getPayload(dataStr2);
+  }
 
+  // in case this is a DKG init or setup request, we must extract the proxy addr
+  if (uri.match(regex)) {
```
nkcr, line 418:
```suggestion
  if (uri.match(regex) || uri.match(dkgInitRegex)) {
```

web/backend/src/Server.ts
cmsigrist: But that won't match a DKG setup request with the '$' at the end will it (since it's an end of string) and the setup still has the {ElectionID} at the end ie. actors/{ElectionID} ?

nkcr: yes you're right, `/\/evoting\/services\/dkg\/actors\/.*$/` would be the one then. Just to be as much explicit as possible.

nkcr (approved):

PR #163, opened 2022-09-05
Updates Dela+Go and fixes the tests

nkcr:
- Uses the latest version of Dela
- Fixes the tests (encoding of ballot ID, warnings instead of errors)
- Uses Go 1.19
- Cleans some parts (removes prints, formats lines)


contracts/evoting/types/ballots_test.go
```diff
@@ -382,14 +383,6 @@ func TestSubject_IsValid(t *testing.T) {
...
 	require.True(t, valid)
 
-	// with wrongly ID not in base64
-	mainSubject.ID = "zzz"
-
-	configuration.Scaffold = []Subject{*mainSubject}
-
-	valid = configuration.IsValid()
-	require.False(t, valid)
-
```
emduc, line 392: Not sure why the test is removed, do we not enforce base64 IDs anymore?

emduc (approved):
Seems all good to me, thanks for the request! :)
It's much clearer with warnings instead of errors indeed.

contracts/evoting/types/ballots_test.go
nkcr: Not sure anymore. Isn't it something you changed based on some errors we had on the frontend ?

emduc: You're right, we no longer require the IDs to be base64 encoded in the configuration, but only when the ballots are cast

nkcr:
Wow, all tests passing. Haven't see that for a while 😄

PR #166, opened 2022-09-13
Fixes initialisation process on the web frontend

### LINKED ISSUE #165: Web frontend should only initialize nodes not already initialized
If for some reasons some nodes are initialized and others not, the "Initialize" action will send the request to all nodes, including those already initialized. There should be a check to send this request only to non-initialized nodes.

nkcr:
Fix #165


web/frontend/src/pages/election/components/utils/useChangeAction.tsx
```diff
@@ -389,8 +389,8 @@ const useChangeAction = (
 
       // TODO: can be modified such that if the majority of the node are
       // initialized than the election status can still be set to initialized
-      const promises = Array.from(nodeProxyAddresses.values()).map((proxy) => {
-        if (proxy !== '') {
+      const promises = Array.from(nodeProxyAddresses).map(([node, proxy]) => {
+        if (proxy !== '' && DKGStatuses.get(node) === NodeStatus.NotInitialized) {
```
cmsigrist, line 393: I don't think this is enough to prevent the initialization of nodes that were already initialized, because of this *cough* beautiful `Promise.all` (I'm sorry .....) line 221, which either set them all up to Initialized or none...

I think one way to do this would be to use the component `DKGStatusRow`, so that each node initializes itself and polls it own status. `useChangeAction` could then be notified of the changes via the states in `DKGStatuses` ?

nkcr: Luckily that works in this case, because "undefined" (and any other non-promise values) count as "resolved". 🙃

> I think one way to do this would be to use the component DKGStatusRow, so that each node initializes itself and polls it own status. useChangeAction could then be notified of the changes via the states in DKGStatuses ?

Yes, that would be much better indeed.

cmsigrist: Yes but it will return `undefined` only if the proxy address of the node could not be found (i.e `proxy===''`). Otherwise in `pollDKG`:
```
// Add a timeout
if (attempts === maxAttempts) {
    throw new Error('Timeout');
}
...
catch (e) {
   return reject(e);
}
```
So if during the poll one of the node times out for some reason the Promise will be rejected thus the `Promise.all` will resolve to rejected for all of them. And the request to initialize will again be sent for all the nodes ...?

jbsv (commented):
Thanks for sharing. However, there is nothing I can add to help here.
🍶 time...

web/frontend/src/pages/election/components/utils/useChangeAction.tsx
nkcr: oh yes, that's true :/

cmsigrist: I can try to solve this this coming week since it will be the start of term, but after that I probably won't have the time anymore 😅

nkcr: I moved the DKG logic to DKGStatusRow, as suggested. That wasn't straightforward as there are probably still too much states spread around between `useChangeAction.tsx`, `Show.tsx`, and `DKGStatusRow.tsx`. Having your input on that would help, if you are willing to :)

nkcr:
The setup process is quite shiny now:


https://user-images.githubusercontent.com/1918021/190392631-4a7a4523-6753-4a7d-a9cb-17f9220b1ce3.mov

web/frontend/src/pages/election/components/DKGStatusRow.tsx
```diff
@@ -50,13 +62,107 @@ const DKGStatusRow: FC<DKGStatusRowProps> = ({
     signal: signal,
   };
 
+  const [showModal, setShowModal] = useState(false);
+
+  // Notify the parent each time our status changes
+  useEffect(() => {
+    notifyDKGState(node, InternalDKGInfo.fromStatus(status));
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [status]);
```
cmsigrist, line 71: If a node fails to initialize once, it get the status Failed. If it fails another time to initialize, then it gets stuck in Initializing, because the status does not change, so the parent are not notified. This normally works:

```suggestion
  }, [DKGLoading]);
```

web/frontend/src/pages/election/components/utils/useChangeAction.tsx
cmsigrist: It's already so much better ! But yes there are way too many states to be passed around, it's a bit cumbersome ....

web/frontend/src/pages/election/components/DKGStatusRow.tsx
```diff
@@ -50,13 +62,107 @@ const DKGStatusRow: FC<DKGStatusRowProps> = ({
...
+      const txt = await response.text();
+      throw new Error(txt);
+    }
+  };
+
+  // Initialize the node if the initialization is ongoing and we are in a
+  // legitimate status.
+  useEffect(() => {
+    if (
+      ongoingAction === OngoingAction.Initializing &&
```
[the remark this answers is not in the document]
cmsigrist: Well it's actually the same problem as before, the parents are not notified since there are no changes the second time we click on Initializing (neither the Status nor the DKGLoading states change...)

web/frontend/src/pages/election/components/utils/useChangeAction.tsx
```diff
@@ -505,18 +393,6 @@ const useChangeAction = (
   };
 
   const getAction = () => {
-    // Check that more than 2/3 of the nodes are working, if not actions are
```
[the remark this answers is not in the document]
nkcr: Not ideal, but I prefer to keep the frontend a bit more simple, and let the user see the error if the action fails (which we don't have here: if the action is not available the user won't know why...)

web/frontend/src/pages/election/components/DKGStatusRow.tsx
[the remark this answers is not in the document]
nkcr: This should be solved in `Show.tsx` and inside the `notifyDKGState` function. There we can count the number of Initialized nodes (even if a node is already initialized, it will call this function at least once with its Initialized status), and update the status+ongoingAction once all nodes are initialized. What do you think ?

web/frontend/src/pages/election/components/DKGStatusRow.tsx
nkcr: True. Then I should manually handle this case and call `notifyDKGState` once I get the second - same - initialization failure.

web/frontend/src/pages/election/components/utils/useChangeAction.tsx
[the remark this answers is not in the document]
cmsigrist: Ok yes I see your point !

web/frontend/src/pages/election/components/DKGStatusRow.tsx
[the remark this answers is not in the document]
cmsigrist: I'm not sure I understood exactly what you meant
There's already a `useEffect` that checks if all the node were initialized and transitions to the Status Initialized if so (in `Show.tsx` line 165). But it's not triggered because none of the entries in `DKGStatuses` change. 
So yes if a node is already initialized, I think it should still call `notifyDKGState` so that it doesn't stay stuck in Initializing.

But I have no idea how it doesn't always remain stuck in Initializing because the `setOngoingAction(ongoingAction.None)` used to be handled in `useChangeAction` with the `onFulfilled/onReject` functions but is now missing for the case of the status Initialized... 😂

web/frontend/src/pages/election/components/DKGStatusRow.tsx
cmsigrist: Yep !

web/frontend/src/pages/election/Show.tsx
```diff
@@ -141,14 +169,20 @@ const ElectionShow: FC = () => {
```
cmsigrist, line 169: I'm not sure but this might be missing, since it's not set to `None` anywhere even if it successfully initializes ?
```suggestion
          setOngoingAction(OngoingAction.None);
```

web/frontend/src/pages/election/components/DKGStatusRow.tsx
nkcr: Even if the state doesn't change, the node notifies the parent with the `notifyLoading` callback. So the parent can still make the check.

web/frontend/src/pages/election/Show.tsx
nkcr: correct !

web/frontend/src/pages/election/components/DKGStatusRow.tsx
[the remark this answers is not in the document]
nkcr: I fixed that with the `DKGLoading`. Even if the status doesn't change, the node will always update its `DKGLoading` and therefore notify the parent with `notifyLoading`. The parent at that time knows that something changed on its children, and can make the necessary checks.

services/dkg/pedersen/handler.go
```diff
@@ -255,19 +260,23 @@ func (h *Handler) start(start types.Start, deals, resps *list.List, from mino.Ad
 // doDKG calls the subsequent DKG steps
 func (h *Handler) doDKG(deals, resps *list.List, out mino.Sender, from mino.Address) {
 	h.log.Info().Str("action", "deal").Msg("new state")
+	*h.status = dkg.Status{Status: dkg.Dealing}
```
pierluca, line 263: If I understand correctly, this is unsafe, because while we only start `Stream` once, we could still receive multiple `types.Start` message. If so, this will be called multiple times.
I believe we have the same issue in dela.

Concretely, the impact on this line is minimal, but I'm concerned about the multiple starts in parallel 🤔 am I missing something ? do we have a test that verifies that we'll ignore the second start (on the same stream) if one is already running ?

web/frontend/src/pages/election/Show.tsx
```diff
@@ -124,10 +148,14 @@ const ElectionShow: FC = () => {
...
   }, [roster]);
 
+  // Keep the "DKGLoading" state according to "nodeLoading". This state tells if
+  // one of the element on the map is true.
   useEffect(() => {
     if (nodeLoading !== null) {
       if (!Array.from(nodeLoading.values()).includes(true)) {
         setDKGLoading(false);
+      } else {
+        setDKGLoading(true);
```
pierluca, line 158:
```suggestion
      const someNodeLoading = Array.from(nodeLoading.values()).includes(true);
      setDKGLoading(someNodeLoading);
```

web/frontend/src/pages/election/Show.tsx
```diff
@@ -124,10 +148,14 @@ const ElectionShow: FC = () => {
...
       }
     }
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
pierluca, line 161: in general exhaustive deps are a very good thing, shouldn't these issues be addressed ?
ignoring them has cost me a lot of time bug fixing, in the past :-/

pierluca (commented):
I'm too far from this code base to be of much help, but I had a quick look and it seems a *lot*  of state is just passed around between components. As per our (offline) discussion @nkcr, I'm pretty sure introducing something like Redux or Akita could provide significant benefits.

web/frontend/src/pages/election/Show.tsx
cmsigrist: We actually tried a couple of times to include all the dependencies with @badrlarhdir during the semester. But for some reasons, the useEffects were not triggered at the correct/expected moment when doing so... 🤔

web/frontend/src/pages/election/components/DKGStatusRow.tsx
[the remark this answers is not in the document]
cmsigrist: Since we're allowing Unreachable nodes to still call the function, we need to make sure the proxy address at least could be retrieved (to avoid a request with an empty proxy address)

```suggestion
      ongoingAction === OngoingAction.Initializing && proxy !== '' &&
```

services/dkg/pedersen/handler.go
nkcr: That can technically happen if other nodes participating in the setup decide to **act badly** and send start messages.

web/frontend/src/pages/election/components/DKGStatusRow.tsx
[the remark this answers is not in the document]
nkcr: @cmsigrist what about (added `if (proxy === '')`:

```ts
  useEffect(() => {
    if (
      ongoingAction === OngoingAction.Initializing &&
      (status === NodeStatus.NotInitialized ||
        status === NodeStatus.Failed ||
        status === NodeStatus.Unreachable)
    ) {
      setDKGLoading(true);

      if (proxy === '') {
        setDKGLoading(false);
        setInfo('proxy empty');
        return;
      }

      initializeNode()
        .then(() => {
          setInfo('');
          setStatus(NodeStatus.Initialized);
        })
        .catch((e: Error) => {
          setInfo(e.toString());
          setStatus(NodeStatus.Failed);
        })
        .finally(() => {
          setDKGLoading(false);
        });
    }
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, [ongoingAction]);
```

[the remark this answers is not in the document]
cmsigrist: This works for the initialization (at least it passed the couple of failure cases I ran)  !  🎉

And then for the setup, I'm not sure if we're supposed to be able to retry setting up a node that has already failed the setup, but if this is the case, then it gets stuck in SettingUp.
It executes the `onrejected` part as expected (line 166):

```
      pollDKGStatus(expectedStatus)
        .then(
          () => {},
          (e: Error) => {
            setStatus(NodeStatus.Failed);
            setInfo(e.toString());
          }
        )
        .finally(() => setDKGLoading(false));
```
But it doesn't call `notifyDKGState`.

And then there's a minor bug where if we set up a first node and it fails but then set up another one which succeeds, the error message remains (the red little (!) next to the node status)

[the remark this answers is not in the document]
nkcr: > This works for the initialization (at least it passed all the failure cases I ran) ! 🎉

Yay 🥳

> I'm not sure if we're supposed to be able to retry setting up a node that has already failed the setup, but if this is the case

Yes, we should. 

> But it doesn't call notifyDKGState.

It should, when doing `setStatus(NodeStatus.Failed);`, then the `useEffect` in line 68 calls `notifyDKGState`, at least if the state was updated. Was it what you meant ?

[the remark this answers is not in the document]
cmsigrist: Yes it should, but I don't think it actually does, otherwise it wouldn't be stuck in Setting up ?

nkcr:
> I'm too far from this code base to be of much help, but I had a quick look and it seems a _lot_ of state is just passed around between components. As per our (offline) discussion @nkcr, I'm pretty sure introducing something like Redux or Akita could provide significant benefits.
> 
> I say this because then it would allow big chunks of the logic to be moved into dedicated services, dealing only with the network and the state, whereas the components would focus on showing what's in the state and triggering the services. If the state is external, it can just be queried as needed.

Instead of pulling out the big boys, an interesting incremental step could be to use the native React context API to wrap all those properties.

web/frontend/src/pages/election/components/DKGStatusRow.tsx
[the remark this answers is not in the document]
nkcr: True. I simplified some elements on `Show.ts` and it seems to work now ... 🥵

web/frontend/src/pages/election/Show.tsx
```diff
@@ -206,7 +235,7 @@ const ElectionShow: FC = () => {
             <div className="font-bold uppercase text-lg text-gray-700 pb-2">{t('action')}</div>
             <div className="px-2">
               {DKGLoading && <LoadingButton>{t('actionLoading')}</LoadingButton>}{' '}
```
cmsigrist, line 237: You forgot about this one 😬
``DKGLoading && ongoingAction === OngoingAction.None &&``

cmsigrist (commented):
Woohoo it works ! 🎉

cmsigrist (approved):

nkcr:
> Woohoo it works ! 🎉

Yay 🥳
Thank you very much for the review, I owe you one 🍺

cmsigrist:
Hahaha no problem (we were kind of fixing the mess I left anyway 😂) !

PR #171, opened 2022-09-21
Adds a CLI to init and setup DKG

pierluca:


pierluca:
This was just for demo purposes. Closing.

PR #174, opened 2022-09-23
Update README for first us of backend and add Troubleshoot section

alberttkt:
-Added a new command to set the backend for the first time and 
-Solution for a possible error during the setup


nkcr (commented):
Very minor comments, fine otherwise :)

nkcr (approved):

PR #177, opened 2022-10-02
Add types in index.tsx

Khadija21102:


web/frontend/src/index.tsx
```diff
@@ -69,18 +69,25 @@ class FlashMessage {
     return this.level;
   }
 }
-
+const flashM = new FlashMessage('', 1);
+const defaultFlashState = {
+  getMessages: function (): FlashMessage[] {
+    return [flashM];
+  },
+  addMessage: function (msg: '', level: 0): void {},
```
nkcr, line 77: not sure about that though, but I would expect to have the types instead of default values:

```suggestion
  addMessage: function (msg: string, level: number): void {},
```

nkcr (approved):
Thanks, LGTM 🚀

PR #183, opened 2022-10-07
Refactoring in cli/cosipbftcontroller

alberttkt:
create a constant for "injector: %v"

PR #184, opened 2022-10-08
Refactoring of contracts/evoting/types/ballots_test.go :

mamine2207:
Defined constants for the most duplicated strings (select, rank, text and the string used when an answer could not be marshaled) to reduce code smells.

### DIFF
--- a/contracts/evoting/types/ballots.go
+++ b/contracts/evoting/types/ballots.go
@@ -278,7 +278,6 @@ type Subject struct {
 // GetQuestion finds the question associated to a given ID and returns it
 // Returns nil if no question found.
 func (s *Subject) GetQuestion(ID ID) Question {
-	// Iterating through the subjects and calling the GetQuestion method on each subject.
 	for _, subject := range s.Subjects {
 		question := subject.GetQuestion(ID)
 		if question != nil {
@@ -324,7 +323,6 @@ func (s *Subject) MaxEncodedSize() int {
 		size += len(rank.Choices) * 4
 	}
 
-	// Calculating the size of the data structure.
 	for _, selection := range s.Selects {
 		size += len(selection.GetString() + "::")
 		size += len(selection.ID)
@@ -427,6 +425,7 @@ type Select struct {
 	Choices []string
 }
 
+// GetString implements Question
 func (s Select) GetString() string {
 	return selectStr
 }
--- a/contracts/evoting/types/ballots.go
+++ b/contracts/evoting/types/ballots.go
@@ -10,6 +10,12 @@ import (
 	"golang.org/x/xerrors"
 )
 
+const (
+	selectStr = "select"
+	rankStr   = "rank"
+	textStr   = "text"
+)
+
 // Ballot contains all information about a simple ballot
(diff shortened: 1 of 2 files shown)


contracts/evoting/types/ballots_test.go
```diff
@@ -8,17 +8,23 @@ import (
...
-
-var ballot2 = string("select:" + encodedQuestionID(1) + ":0,0,0\n" +
-	"rank:" + encodedQuestionID(2) + ":128,128,128,128\n" +
-	"select:" + encodedQuestionID(3) + ":0,0,0,0,0\n" +
-	"text:" + encodedQuestionID(4) + ":xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx," +
+const SELECT_STR = "select:"
+const RANK_STR = "rank:"
+const TEXT_STR = "text:"
+const UNMARSHALING_RANK_STR = "could not unmarshal rank answers: "
+const UNMARSHALING_TEXT_STR = "could not unmarshal text answers: "
```
emduc, line 15: We use camel case for constants as well. You can also declare them altogether since they're related, see PR #183 or `contract/evoting.go` for examples

emduc (changes requested):
Nice ! Just need a small change on the naming convention :)

Food for thoughts: would it make sense to have those strings linked to the questions and accessible with a getter?

contracts/evoting/types/ballots_test.go
mamine2207: yes, I think it would make sense and we will be able to use them in some functions in ballots.go such as unmarshal(...) or MaxEncodedSize(...). The only problem I can see is that the format is slightly different between the strings in the tests and the ones in ballots.go and that they are mostly recurrent in the tests. 
Thanks for the note on the camel case convention, I have to get rid of my Java habits :)

emduc: Do you mean because of the `:` ? Since this is just formatting it does not have to be part of the string. It could be written as a join in the tests.
This might be overthinking it (although quite clean), but we could use the Question interface to force the Text/Rank/Select questions to re-define the "String()" function which would simply return "select"/"rank"/"text" and could be called everywhere. It's just an idea though ;)

contracts/evoting/types/ballots.go
```diff
@@ -401,6 +407,7 @@ type Question interface {
 	GetMaxN() uint
 	GetMinN() uint
 	GetChoicesLength() int
+	GetString() string
```
nkcr, line 410: In this case `GetID() string` feels more appropriate.

nkcr: Then, you can rename `selectStr` to `selectID`, same for rank and text.

nkcr (changes requested):

contracts/evoting/types/ballots.go
mamine2207: Ok, same thing for the constants in the test file, right ?

nkcr (approved):
LGTM 🚀

PR #186, opened 2022-10-11
Add Contributors to the About page

aelalamy42:
This patch will add a section at the end of the about page containing all the current contributors. I also added a "contributors" field for later i18n and translation.


aelalamy42:
Really good point, it was not useful at all. I removed it :)

nkcr (approved):

PR #192, opened 2022-10-17
configuration SonarCloud

alberttkt:
added SonarCloud configuration so that tests are not included in coverage tests and code smells


emduc (approved):

emduc:
Nice ! Just to be sure I understand this right, the multicriteria lines are to ignore naming warnings on tests?

nkcr (commented):
Thanks!

I have same question as Emilien though, what's this `multicriteria` ? 😅

alberttkt:
I found it on SonarCloud it seems to be to exclude a given analysis for the files path you choose:
image

PR #201, opened 2022-11-01
Readjustement of the introduction image and first version of casbin authorization mechanism

Ghita2002:


web/backend/model.conf
```diff
@@ -0,0 +1,11 @@
...
+r = sub, obj, act
+
+[policy_definition]
+p = sub, obj, act
+
+[policy_effect]
+e = some(where (p.eft == allow))
+
+[matchers]
+m = r.sub == p.sub && r.obj == p.obj && r.act == p.act
```
pierluca, line 11: what this file does is not immediately obvious, can you add some documentation (ideally, inline) to indicate what each line does ?

web/backend/src/Server.ts
```diff
@@ -175,78 +176,98 @@ app.get('/api/personal_info', (req, res) => {
...
+// return false;
+// }
 
-  const { role } = req.session;
+// const { role } = req.session;
 
-  return roles.includes(role as string);
-}
+// return roles.includes(role as string);
+// }
```
pierluca, line 188: feel free to delete code that is not relevant anymore. Commented out code is generally frowned upon 😄

web/backend/src/Server.ts
```diff
@@ -175,78 +176,98 @@ app.get('/api/personal_info', (req, res) => {
...
-    res.status(400).send('Unauthorized - only admins allowed');
-    return;
-  }
-
-  const opts: RangeOptions = {};
-  const users = Array.from(
-    usersDB.getRange(opts).map(({ key, value }) => ({ id: '0', sciper: key, role: value }))
-  );
-  res.json(users);
+app.get('/api/user_rights', async (req, res) => {
```
pierluca, line 197: are you sure that express supports async handlers natively?

web/backend/src/Server.ts
```diff
@@ -175,78 +176,98 @@ app.get('/api/personal_info', (req, res) => {
...
+  e.then((enforcer) =>
```
pierluca, line 198: if you have an async function, you don't need to do `.then` and `.catch`, you can just use `await` where you would use a then, and `try/catch` instead of `.catch`.

see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await

web/backend/src/Server.ts
```diff
@@ -175,78 +176,98 @@ app.get('/api/personal_info', (req, res) => {
...
+          const users = Array.from(
+            usersDB.getRange(opts).map(({ key, value }) => ({ id: '0', sciper: key, role: value }))
+          );
+          res.json(users);
+        } else {
+          res.status(400).send('Unauthorized - only admins allowed');
+        }
+      })
+      .catch((erreur) => console.log('error', erreur))
+  ).catch((erreur1) => console.log('error', erreur1));
```
pierluca, line 213: avoid french words in the codebase 😁 
also, you can reuse the same variable name, as it's scoped just to that function.

pierluca (commented):
Nice progress ! 
I just made a quick pass on the codebase, this is not a full in-depth review and I'll let @emduc  and @nkcr give their own feedback.

There are a couple of remarks that apply throughout. Let me know if you have questions 😄

web/backend/src/Server.ts
```diff
@@ -26,6 +26,27 @@ const app = express();
...
 
+let enf: Enforcer;
+
+const enforcerLoading = newEnforcer('model.conf', 'policy.csv');
+const e = newEnforcer('model.conf', 'policy.csv');
+const port = process.env.PORT || 5000;
+
+Promise.all([enforcerLoading])
+  .then((res) => {
+    [enf] = res;
```
pierluca, line 37: Do you need a `Promise.all` if you're looking at just one promise resolving?
Why can't you do the following ? 
```suggestion
enforcerLoading
  .then((enf) => {
```
I may be missing something here, so do let me know if I'm wrong 😁

web/backend/src/Server.ts
```diff
@@ -26,6 +26,27 @@ const app = express();
 
 app.use(morgan('tiny'));
 
+let enf: Enforcer;
+
+const enforcerLoading = newEnforcer('model.conf', 'policy.csv');
+const e = newEnforcer('model.conf', 'policy.csv');
```
pierluca, line 32: this looks unused

web/backend/src/Server.ts
```diff
@@ -26,6 +26,27 @@ const app = express();
 
 app.use(morgan('tiny'));
 
+let enf: Enforcer;
+
+const enforcerLoading = newEnforcer('model.conf', 'policy.csv');
```
pierluca, line 31: do you need this to be in the global scope, or could you put it (along with the promise resolution below) in an initialization function ?

web/backend/src/Server.ts
```diff
@@ -224,13 +245,12 @@ app.post('/api/add_role', (req, res) => {
 
 // This call (only for admins) allow an admin to remove a role to a user.
 app.post('/api/remove_role', (req, res) => {
-  if (!isAuthorized(['admin'], req)) {
+  if (!isAuthorized(req.session.userid, 'database', 'removerole')) {
```
pierluca, line 248: something I forgot to mention, you might want to create string constants for the `'election'`, `'create'` and the other enforcement parameters. This way, it'll be harder for someone to make a typo 😁

web/backend/src/Server.ts
```diff
@@ -461,13 +475,18 @@ function sendToDela(dataStr: string, req: express.Request, res: express.Response
...
 // Secure /api/evoting to admins and operators
 app.use('/api/evoting/*', (req, res, next) => {
-  if (!isAuthorized(['admin', 'operator'], req)) {
-    console.log('role is:', req.session.role);
-    res.status(400).send('Unauthorized - only admins and operators allowed');
-    return;
-  }
-
-  next();
+  e.then((enforcer) => {
```
pierluca, line 478: is there are a reason not to use `isAuthorized` here?

web/backend/src/Server.ts
```diff
@@ -461,13 +475,18 @@ function sendToDela(dataStr: string, req: express.Request, res: express.Response
...
+      .enforce(req.session.userid, 'election', 'create')
+      .then((isOk) => {
+        if (isOk) {
+          next();
+        } else {
+          res.status(400).send('Unauthorized - only admins and operators allowed');
+        }
+      })
+      .catch((error) => console.log('error', error));
+  }).catch((error1) => console.log('error', error1));
```
pierluca, line 489: As per our discussion today, this should work:
```suggestion
  e.then((enforcer) => {
    return enforcer
      .enforce(req.session.userid, 'election', 'create')
      .then((isOk) => {
        if (isOk) {
          next();
        } else {
          res.status(400).send('Unauthorized - only admins and operators allowed');
        }
      })
  }).catch((error) => console.log('error', error));
```

No need to do a double-catch 😁 
And since you're directly returning a value, you could also do this (notice the absence of `{}` and `return`):

```suggestion
  e.then((enforcer) =>
    enforcer
      .enforce(req.session.userid, 'election', 'create')
      .then((isOk) => {
        if (isOk) {
          next();
        } else {
          res.status(400).send('Unauthorized - only admins and operators allowed');
        }
      })).catch((error) => console.log('error', error));
```

web/backend/src/Server.ts
```diff
@@ -558,7 +577,7 @@ app.get('*', (req, res) => {
   res.status(404).send(`not found ${xss(url.toString())}`);
 });
 
-const port = process.env.PORT || 5000;
-app.listen(port);
+// const port = process.env.PORT || 5000;
+// app.listen(port);
 
-console.log(`App is listening on port ${port}`);
+// console.log(`App is listening on port ${port}`);
```
pierluca, line 583: feel free to delete code, no need to comment it out 😁

web/backend/model.conf
Ghita2002: I added some quick comments.

web/backend/src/Server.ts
Ghita2002: I modify that using the proposition of Noemien.

web/backend/src/Server.ts
Ghita2002: Thanks for the view, it is fixed.

web/backend/src/Server.ts
Ghita2002: I just forget it. It is fixed !

web/backend/src/Server.ts
```diff
@@ -177,46 +186,49 @@ app.post('/api/logout', (req, res) => {
...
-      islogged: false,
-    });
-  }
+  const m = new Map<String, Array<String>>();
+  enf.getFilteredPolicy(0, String(req.session.userid)).then((list) => {
+    for (let i = 0; i < list.length; i += 1) {
+      for (let j = 1; j < list[0].length; j += +2) {
+        console.log(list[i][j]);
+        if (m.has(list[i][j])) {
+          m.get(list[i][j])?.push(list[i][j + 1]);
```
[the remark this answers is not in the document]
Ghita2002: Even if I check `m.has` I still have the error (Object is possibly 'undefined') This why I used `?`. However it can be a good way to test the result of `m.get`. I was just wondering how it is better than using `?`

web/backend/src/Server.ts
```diff
@@ -154,49 +185,58 @@ app.post('/api/logout', (req, res) => {
 // As the user is logged on the app via this express but must also be logged in
 // the react. This endpoint serves to send to the client (actually to react)
 // the information of the current user.
```
nkcr, line 187: The comment is for the statement bellow, not the `setMapAuthorization` function, which must also have its own comment describing what it does.

web/backend/src/Server.ts
```diff
@@ -154,49 +185,58 @@ app.post('/api/logout', (req, res) => {
...
-  }
-});
-
-function isAuthorized(roles: string[], req: express.Request): boolean {
-  if (!req.session || !req.session.userid) {
-    return false;
+function setMapAuthorization(list: string[][]) {
+  const m = new Map<String, Array<String>>();
+  for (let i = 0; i < list.length; i += 1) {
+    for (let j = 1; j < list[0].length; j += +2) {
```
nkcr, line 191: you should explain (1) what is contained in `list[0]`, and (2) why you are doing `+= 2`.

web/backend/src/Server.ts
```diff
@@ -154,49 +185,58 @@ app.post('/api/logout', (req, res) => {
...
+      console.log(list[i][j]);
```
nkcr, line 192: you probably used it for debugging. To be removed.

web/backend/src/Server.ts
nkcr: This function is hard to understand without any explanation. You should at least describe what contains `list: string[][])`.

web/backend/src/Server.ts
```diff
@@ -154,49 +185,58 @@ app.post('/api/logout', (req, res) => {
...
-  const { role } = req.session;
-
-  return roles.includes(role as string);
+  console.log(m);
+  return m;
 }
+app.get('/api/personal_info', (req, res) => {
+  let m = new Map<String, Array<String>>();
+  enf.getFilteredPolicy(0, String(req.session.userid)).then((list) => {
+    m = setMapAuthorization(list);
```
nkcr, line 206: If the user is not logged you can avoid calling `setMapAuthorization` and return an empty map.

web/backend/src/Server.ts
```diff
@@ -8,9 +8,19 @@ import crypto from 'crypto';
 import lmdb, { RangeOptions } from 'lmdb';
 import xss from 'xss';
 import createMemoryStore from 'memorystore';
+import { Enforcer, newEnforcer } from 'casbin';
 
 const MemoryStore = createMemoryStore(session);
-
+const ROLES = 'roles';
```
nkcr, line 14: This is way too generic. You should prefix the variable name to indicate how the variable is used:
```
SUBJECT_ROLES = 'roles';
SUBJECT_PROXIES = 'proxies';
...
ACTION_LIST = 'list';
ACTION_REMOVE = 'remove';
...
```

web/frontend/src/layout/NavBar.tsx
```diff
@@ -24,6 +24,17 @@ import { Popover, Transition } from '@headlessui/react';
 import { LoginIcon, LogoutIcon, MenuIcon, XIcon } from '@heroicons/react/outline';
 import { PlusIcon } from '@heroicons/react/solid';
 
+const ELECTION = 'election';
```
nkcr, line 27: You should prefix the variable name:
```
SUBJECT_ELECTION = ...
...
ACTION_CREATE = ...
```

web/frontend/src/layout/NavBar.tsx
```diff
@@ -24,6 +24,17 @@ import { Popover, Transition } from '@headlessui/react';
...
+const CREATE = 'create';
+const ROLES = 'roles';
+const ADD = 'add';
+const LIST = 'list';
+
+function hasAuthorization(authCtx, subject: string, action: string): boolean {
+  return (
+    authCtx.authorization.has(subject) && authCtx.authorization.get(subject).indexOf(action) > -1
```
nkcr, line 35:
```suggestion
    authCtx.authorization.has(subject) && authCtx.authorization.get(subject).indexOf(action) !== -1
```

nkcr (changes requested):

web/backend/src/Server.ts
Ghita2002: I added some comments and while doing it I noticed that the iteration on j isn't necessary.

web/backend/src/Server.ts
Ghita2002: I delete that.

web/backend/src/Server.ts
Ghita2002: Fixed !

web/backend/src/Server.ts
```diff
@@ -182,23 +183,30 @@ app.post('/api/logout', (req, res) => {
   });
 });
 
+// This function helps us convert the double list of the authorization
+// returned by the casbin function getFilteredPolicy to a map that link
+// an object to the action authorized
+// list[0] contains the policies so list[i][0] is the sciper
+// list[i][1] is the object and list[i][2] is the action
+function setMapAuthorization(list: string[][]) {
+  const m = new Map<String, Array<String>>();
```
pierluca, line 192: do you want to use `String` the object or just `string` the plain data type?

web/backend/src/Server.ts
```diff
@@ -151,52 +183,63 @@ app.post('/api/logout', (req, res) => {
...
+  for (let i = 0; i < list.length; i += 1) {
+    if (m.has(list[i][1])) {
```
nkcr, line 194: The comment is already a great improvement. 👍 

Now, the next step is self-explanatory code.. ;) 

You could do:

```
const [subject, action] = list[i][1], list[i][2];
if (m.has(subject)) {
...
```

web/backend/src/Server.ts
```diff
@@ -151,52 +183,63 @@ app.post('/api/logout', (req, res) => {
...
+      m.get(list[i][1])?.push(list[i][2]);
```
nkcr, line 195: I'm not sure I understand the `?`. Is it to prevent adding multiple times the same action ?

web/backend/src/Server.ts
```diff
@@ -182,23 +183,30 @@ app.post('/api/logout', (req, res) => {
   });
 });
 
+// This function helps us convert the double list of the authorization
+// returned by the casbin function getFilteredPolicy to a map that link
+// an object to the action authorized
+// list[0] contains the policies so list[i][0] is the sciper
+// list[i][1] is the object and list[i][2] is the action
+function setMapAuthorization(list: string[][]) {
```
pierluca, line 191: in typescript it's good to explicitly indicate the return type.

```suggestion
function setMapAuthorization(list: string[][]): Map<string, Array<string>> {
```

web/backend/src/Server.ts
```diff
@@ -207,7 +215,7 @@ app.get('/api/personal_info', (req, res) => {
         firstname: req.session.firstname,
         role: req.session.role,
         islogged: true,
-        authorization: Object.fromEntries(m),
+        authorization: Object.fromEntries(setMapAuthorization(list)),
```
pierluca, line 218: much cleaner :)

web/backend/src/Server.ts
```diff
@@ -182,23 +183,30 @@ app.post('/api/logout', (req, res) => {
...
+  }
+  console.log(m);
+  return m;
+}
+
 // As the user is logged on the app via this express but must also be logged in
 // the react. This endpoint serves to send to the client (actually to react)
 // the information of the current user.
 app.get('/api/personal_info', (req, res) => {
   const m = new Map<String, Array<String>>();
```
pierluca, line 208: do you still need this variable?

pierluca (commented):
Not absolutely perfect, but sooooo much better, **great job** ! 👍 

I've some minor remarks, but at this point it looks mergeable to me :)

web/backend/src/Server.ts
```diff
@@ -151,52 +183,63 @@ app.post('/api/logout', (req, res) => {
...
+        authorization: Object.fromEntries(setMapAuthorization(list)),
+      });
+    } else {
+      res.json({
+        sciper: 0,
+        lastname: '',
+        firstname: '',
+        role: '',
+        islogged: false,
+        authorization: Object.fromEntries(m),
```
nkcr, line 227: I think you can just use `authorization: new Map<String, Array<String>>(),`
or even just `authorization: {},`

web/backend/src/Server.ts
Ghita2002: When searching which one use, I found that it is better to avoid bugs to use `string`

web/backend/src/Server.ts
Ghita2002: I decide to delete it

web/backend/src/Server.ts
Ghita2002: Fixed !

web/backend/src/Server.ts
Ghita2002: It's necessary, however I check that `m.has(list[i][1]`, to make sure that `m.get(list[i][1]` is not undefined.

web/backend/src/Server.ts
Ghita2002: Thanks for the review ! Done

web/backend/src/Server.ts
Ghita2002: Done !

nkcr (approved):
Good to go 🚀

PR #204, opened 2022-11-07
Uses enforceSync to check authorizations

nkcr:
@Ghita2002 this is to show you how you can refactor the authorization check. You can merge this PR, which targets your branch, and update the rest of the code accordingly (I only updated the `/api/user_rights'` handler). 

- Updates the initialisation of Casbin so that we start only when the enforcer is ready
- Uses enforceSync, which makes the checking mechanism was easier
- Adds a utility function isAuthorized

### DIFF
--- a/web/backend/src/Server.ts
+++ b/web/backend/src/Server.ts
@@ -8,7 +8,7 @@ import crypto from 'crypto';
 import lmdb, { RangeOptions } from 'lmdb';
 import xss from 'xss';
 import createMemoryStore from 'memorystore';
-import { newEnforcer } from 'casbin';
+import { Enforcer, newEnforcer } from 'casbin';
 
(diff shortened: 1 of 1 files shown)


web/backend/src/Server.ts
```diff
@@ -187,30 +208,22 @@ app.get('/api/personal_info', (req, res) => {
...
-          );
-          res.json(users);
-        } else {
-          res.status(400).send('Unauthorized - only admins allowed');
-        }
-      })
-      .catch((error) => console.log('error', error))
-  ).catch((error) => console.log('error', error));
+  if (!isAuthorized(req.session.userid, 'roles', 'list')) {
+    res.status(400).send('Unauthorized - only admins allowed');
```
nkcr, line 218: should be an error `401`

PR #206, opened 2022-11-09
Sciper checking

mamine2207:


web/backend/src/Server.ts
```diff
@@ -216,10 +216,23 @@ app.post('/api/add_role', (req, res) => {
...
+      } else {
+        usersDB.put(sciper, role).catch((error) => {
+          res.status(500).send('Failed to add role');
+          console.log(error);
+        });
+      }
+    });
+  } else {
+    res.status(400).send('sciper length incorrect');
+  }
```
pierluca, line 235: In general, you can reduce indentation by using guard clauses and making the code more readable.
Just a suggestion:

```suggestion
  // The sciper has to contain 6 numbers
  if (sciper > 999999 || sciper < 100000) {
    res.status(400).send('sciper length incorrect');
    return;
  }
  
  // call https://search-api.epfl.ch/api/ldap?q=228271 
  // if the answer is empty then the sciper is invalid
  axios.get(`https://search-api.epfl.ch/api/ldap?q=${sciper}`).then((response) => {
    if (response.data.length === 0) {
      res.status(400).send('Unknown sciper');
    } else {
      usersDB.put(sciper, role).catch((error) => {
        res.status(500).send('Failed to add role');
        console.log(error);
      });
    }
  });
```

web/backend/src/Server.ts
```diff
@@ -216,10 +216,23 @@ app.post('/api/add_role', (req, res) => {
...
 
-  usersDB.put(sciper, role).catch((error) => {
-    res.status(500).send('Failed to add role');
-    console.log(error);
-  });
+  // The sciper has to contain 6 numbers
+  if (sciper < 999999 && sciper > 100000) {
+    // call https://search-api.epfl.ch/api/ldap?q=228271 if the answer
+    // empty then sciper invalid
+    axios.get(`https://search-api.epfl.ch/api/ldap?q=${sciper}`).then((response) => {
```
pierluca, line 223: what if the `get` fails ? it seems you don't catch that case.

web/backend/src/Server.ts
```diff
@@ -216,10 +216,23 @@ app.post('/api/add_role', (req, res) => {
...
+      if (response.data.length === 0) {
+        res.status(400).send('Unknown sciper');
+      } else {
+        usersDB.put(sciper, role).catch((error) => {
+          res.status(500).send('Failed to add role');
+          console.log(error);
+        });
+      }
+    });
```
pierluca, line 232: For you information, you can also concatenate promises, by returning a promise from within the body of a `then`.

This is sometimes more readable, and you can group all the errors in one catch.
```suggestion
    axios.get(`https://search-api.epfl.ch/api/ldap?q=${sciper}`)
      .then((response) => {
        if (response.data.length === 0) {
          res.status(400).send('Unknown sciper');
          return;
        }
        
        return usersDB.put(sciper, role);
      })
      .catch((error) => {
        res.status(500).send('Failed to add role');
        console.log(error);
      });;
```

pierluca (commented):
Nice little PR. I've added a few suggestions for improvement.

web/backend/src/Server.ts
mamine2207: Yes I missed this one, I just added a catch in a new commit

web/backend/src/Server.ts
```diff
@@ -216,10 +216,33 @@ app.post('/api/add_role', (req, res) => {
...
+        res.status(400).send('Unknown Sciper');
+        return;
+      }
+
+      usersDB.put(sciper, role).catch((error) => {
+        res.status(500).send('Failed to add role');
+        console.log(error);
+      });
+
+      res.status(200).send('Role added');
```
pierluca, line 240: this will also execute in the case of the `catch` above, don't you want to put in a `then` (before the `catch`) ?

PR #208, opened 2022-11-13
Created new integration tests

alberttkt:


pierluca:
Would the tests work with a timeout higher than 7 min? :)

integration/dvotingdela.go
```diff
@@ -137,7 +137,11 @@ func setupDVotingNodes(t require.TestingT, numberOfNodes int, tempDir string) []
 		delaNodes = append(delaNodes, node)
 		dVotingNodes = append(dVotingNodes, node)
 	}
-	delaNodes[0].Setup(delaNodes[1:]...)
+	if node != nil {
+		node.Setup(delaNodes...)
```
pierluca, line 141: the `node` name is a bit confusing, given there's a variable right above also called `node` which has a different meaning and values.

integration/performance_test.go
```diff
@@ -46,7 +46,7 @@ func BenchmarkIntegration_CustomVotesScenario(b *testing.B) {
 	b.Logf("using temp dir %s", dirPath)
 
 	// ##### CREATE NODES #####
-	nodes := setupDVotingNodes(b, numNodes, dirPath)
+	nodes := setupDVotingNodes(b, numNodes, dirPath,nil)
```
pierluca, line 49: I suspect you want to run `go fmt` :-) 

```suggestion
	nodes := setupDVotingNodes(b, numNodes, dirPath, nil)
```

pierluca (commented):
It's usually best to avoid files that are more than a few hundred lines long.
It looks like the integration tests are now ready to be split across multiple files.
What do you think?

I'm suggesting this before other comments, so that we could comment directly on the right files.

integration/integration_test.go
```diff
@@ -44,7 +44,29 @@ var serdecontext = json.NewContext()
...
+
+func TestAddNodes(t *testing.T) {
+	t.Run("5 nodes, 10 votes add 2 node during the process", getIntegrationTestAddNodes(5, 10, 2))
+}
+
+func BenchmarkIntegration(b *testing.B) {
+	b.Run("10 nodes, 100 votes", getIntegrationBenchmark(10, 100))
 }
 
 func getIntegrationTest(numNodes, numVotes int) func(*testing.T) {
```
emduc, line 72: Random idea: You could have the # of bad votes, revotes,  crash and new nodes added as optional arguments to avoid duplicating this function for each purpose (and allow for combinations)

integration/integration_test.go
```diff
@@ -67,7 +89,598 @@ func getIntegrationTest(numNodes, numVotes int) func(*testing.T) {
...
+
+		t.Logf("start casting votes")
+		form, err := getForm(formFac, formID, nodes[0].GetOrdering())
+		require.NoError(t, err)
+
+		_, err = castVotesRandomly(m, actor, form, numVotes)
+		require.NoError(t, err)
+
+		castedVotes, err := castVotesRandomly(m, actor, form, numVotes)
+		require.NoError(t, err)
```
emduc, line 416: the number of times votes are re-cast could be an argument or a random number. With the current implementation it doesn't make a difference, but it's always better to define stronger invariants in case we were to change that implementation later.

integration/integration_test.go
```diff
@@ -67,7 +89,598 @@ func getIntegrationTest(numNodes, numVotes int) func(*testing.T) {
...
+
+		var crashNodeList []dVotingCosiDela
+		for i := 0; i < failingNodes; i++ {
+			crashID := rand.Intn(numNodes - i)
+			crashNode := nodes[crashID]
+			nodes = append(nodes[:crashID], nodes[crashID+1:]...)
+			crashNodeList = append(crashNodeList, crashNode)
+		}
+		err = closeNodes(crashNodeList)
+		require.NoError(t, err)
```
emduc, line 565: It would be cool to be able to choose at which stage the nodes fail (for example by replacing the # of crashes by a list of integers where the number at index i represents the number of crash at stage i)

emduc (commented):
Nice ! This will be very useful. Just added a few comments about the tests and how you could make them a bit more thorough :).

integration/utility_functions.go
```diff
@@ -0,0 +1,524 @@
+package integration
```
pierluca, line 1: Just a tiny comment: this file has a massive amount of useful, but disjoint utilities.
I would suggest perhaps splitting them by functionality: you could have a file for form-related operations, one for dkg-related stuff, another for casting votes, etc. 

Ideally, you want to strive for [high cohesion](https://en.wikipedia.org/wiki/Cohesion_(computer_science)) and low coupling at all levels in the software.

alberttkt:
Changes are visible in #231

PR #222, opened 2022-11-23
verify signature before process the request for every signedRequest

Flamewind97:
For every signedRequest,
we will validate the signature before processing other parameters


emduc (approved):
Makes sense :)

nkcr (approved):

PR #223, opened 2022-11-23
add check of rosterLen before sending dkg request

Flamewind97:
Before defines every addrs array in services, we check the rosterLen is not empty to have an early return

### DIFF
--- a/services/dkg/pedersen/mod.go
+++ b/services/dkg/pedersen/mod.go
@@ -213,6 +213,12 @@ func (a *Actor) Setup() (kyber.Point, error) {
 		return nil, err
 	}
 
+	if form.Roster.Len() == 0 {
+		err := xerrors.Errorf("the roster is empty")
+		a.setErr(err, nil)
+		return nil, err
+	}
+
 	addrs := make([]mino.Address, 0, form.Roster.Len())
 	addrIter := form.Roster.AddressIterator()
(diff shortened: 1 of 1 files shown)


services/dkg/pedersen/mod.go
```diff
@@ -364,6 +370,9 @@ func (a *Actor) ComputePubshares() error {
 	}
 
 	players := mino.NewAddresses(a.handler.startRes.GetParticipants()...)
+	if players.Len() == 0 {
+		return xerrors.Errorf("the list of Participants is empty")
+	}
```
emduc, line 375: should we call a.setErr here ? 
This function is called when we want to decrypt the ballots, so I am wondering what we would be able to do / should do in the scenario where players is empty...

emduc (commented):
Nice small optimization :)

services/dkg/pedersen/mod.go
nkcr: yes, true.
@Flamewind97 please add a `setErr` as in other cases.

nkcr (changes requested):

services/dkg/pedersen/mod.go
nkcr: In the end we don't need, I mixed the context.

nkcr (approved):

PR #227, opened 2022-11-28
Change the name of the JSON file that is uploaded

Ghita2002:


web/frontend/src/pages/form/components/FormForm.tsx
```diff
@@ -105,7 +105,7 @@ const FormForm: FC<FormFormProps> = () => {
     const jsonString = `data:text/json;chatset=utf-8,${encodeURIComponent(JSON.stringify(data))}`;
     const link = document.createElement('a');
     link.href = jsonString;
-    link.download = 'form_configuration.json';
+    link.download = MainTitle + '.json';
```
nkcr, line 108: `Maintitle` is a free user input. Is it safe to directly use it as a filename ?

Ghita2002: What do you mean by free user input ?

nkcr: Users can enter anything they want, and we don't sanitize the content.

Ghita2002: It can effectively be an issue. However, I found that using the title of the form for the json file is quite the most natural. Could we fix some rules when entering the title ?

nkcr: I have some ideas in mind, but what would you suggest ?

nkcr: following up on what I mentioned this morning, something like that would do the trick: `replace(/[^a-zA-Z0-9]+/g, "-")`

PR #231, opened 2022-12-08
New transaction system

alberttkt:
First version of the new transaction system:
working fine but tests are not up to date and the code is not clean already

PR #233, opened 2022-12-13
The New Transaction System's tests

mamine2207:


pierluca:
Hi,
It looks as though this PR contains commits from #231 , would it make sense to create a PR that points to the branch [newTransactionSystem](https://github.com/dedis/d-voting/tree/newTransactionSystem) so that we can only see the differences ? It might make it easier to review, as opposed to reviewing everything twice :-)

PR #236, opened 2022-12-14
Fix the bug in the kill_test script

Ghita2002:


nkcr (changes requested):
Thank you for the fix 🙏 
I see there are other non-related fixes/feature. Could you please just include the small fix in the kill_test.sh, and leave the rest for another PR ? :)

Ghita2002:
> Thank you for the fix 🙏 I see there are other non-related fixes/feature. Could you please just include the small fix in the kill_test.sh, and leave the rest for another PR ? :)

PR #254, opened 2022-12-22
Modify mocks ghita & adding database

Ghita2002:


web/backend/src/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    image: postgres
+    restart: always
+    ports:
+      - 5432:5432
+    environment:
+      POSTGRES_USER: 'dvoting'
+      POSTGRES_PASSWORD: 'dvoting'
+    volumes:
+      - ./data:/var/lib/postgresql/data
+      - ./mysql.sql:/docker-entrypoint-initdb.d/init.sql
```
nkcr, line 14: `mysql.sql` is not a good name. You are not using a mySQL database. `migration.sql` would be a better name.

nkcr (changes requested):
Make sure that all (except integration) tests pass.

PR #261, opened 2022-12-26
#219 Duplicate Function GetForm

### LINKED ISSUE #219: Technical Debt - Duplicate function getForm()
Under dkg folder, struct Actor and struct dkgHandler both have the function getForm() and all the implementations are the same. Thus we can remove one of the functions or call one function from another to reduce duplication.

Where:
services/dkg/pedersen/mod.go getForm()
services/dkg/pedersen/handler.go getForm()

Target:
all the unit tests in dkg should pass.
all the functions should be able to work as before

mamine2207:


pierluca:
This PR seems to add unused functions, with no clear usage, we're closing it.

PR #262, opened 2022-12-29
Improve login Redirection

aelalamy42:
This pull requests aims to improve the redirection after login.
Before: Everytime we would login we would be redirected to the main page of the website which was inconvenient.
Now: Using cookies we are always redirected to the page from which we clicked on the login button.


web/frontend/src/pages/session/HandleLogin.tsx
```diff
@@ -7,6 +7,11 @@ const handleLogin = async (fctx: FlashState) => {
   try {
     const res = await fetch(ENDPOINT_GET_TEQ_KEY);
 
+    const d = new Date();
+    d.setTime(d.getTime() + 120000);
+    let expires = 'expires=' + d.toUTCString();
+    document.cookie = 'redirect' + '=' + window.location.pathname + ';' + expires + ';path=/';
```
nkcr, line 13:
```suggestion
    document.cookie = 'redirect=' + window.location.pathname + '; expires=' + expires + '; path=/';
```

web/frontend/src/pages/session/Logged.tsx
```diff
@@ -12,14 +12,30 @@ const Logged: FC = () => {
   const authCtx = useContext(AuthContext);
   const fctx = useContext(FlashContext);
 
+  const getCookie = () => {
+    let name = 'redirect' + '=';
+    let decodedCookie = decodeURIComponent(document.cookie);
```
nkcr, line 17: There is a simpler way to retrieve a cookie, check the official doc: https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie#example_2_get_a_sample_cookie_named_test2

nkcr (changes requested):

nkcr (approved):

PR #264, opened 2022-12-29
BUGFIX : Change the mock data to make it work with hint button + Minor Display Issue

aelalamy42:
A bug was found where the mocks were not working anymore because they were not giving an empty hint button text and made the website crash. This PR aims to fix this bug


aelalamy42:
I also found an issue where when casting a ballot, the hint button would appear near the center of the page whereas when hovering the actual hint text would appear at the total right of the page. I fixed that by assuring that the hint button was aligned at the right instead of in the middle

nkcr (approved):
Thanks 👍

PR #266, opened 2022-12-30
Restrict the MaxLength to be less than 1000

Ghita2002:


web/frontend/src/schema/configurationValidation.ts
```diff
@@ -280,12 +280,18 @@ const textsSchema = yup.object({
...
+        if (MaxLength <= 0) {
           return this.createError({
             path,
             message: `MaxLength should be at least equal to 1 in texts [objectID: ${ID}]`,
           });
         }
+        if (MaxLength > 1000) {
+          return this.createError({
+            path,
+            message: `MaxLength should be less than 1000 in texts [objectID: ${ID}]`,
```
nkcr, line 292:
```suggestion
            message: `MaxLength should not exceed 1000 in texts [objectID: ${ID}]`,
```

nkcr (changes requested):

nkcr (approved):

PR #271, opened 2023-01-15
Enable editing a node

Ghita2002:
It works well in my branch. However, the new transaction system breaks the frontend so we cannot edit a proxy. This is why for now I didn't pull the main to my branch.


nkcr (dismissed):
Looks good 👍
I am surprised the transaction system makes it fail, could you identify the problem ?

Ghita2002:
@nkcr It's a little bit weird. Actually the request to the backend seems to be done correctly. The backend send a response but the frontend never get it so it throws an exception saying that the proxy url is not valid. This error appears currently also on the main so I thought that it is the change of the transaction mechanism that cause it. (Before that it worked correctly)

nkcr (approved):

PR #286, opened 2023-08-02
fix: add missing dependencies

PascalinDe:


lanterno (approved):
LGTM

lanterno:
@PascalinDe for typescript, the package manager introduces lock files. I believe they should be pushed/git-tracked as well. 
Otherwise, the manager won't recognize the updates completely. However, I think it's still nice to build them from within docker.

pierluca (changes requested):
I agree with @lanterno. You need to commit the package-lock.json as well.

PascalinDe:
added package-lock.json files as generated in Docker environment

naturally, that also changed some other versions...

lanterno (approved):

pierluca (approved):
LGTM

PR #288, opened 2023-08-02
feat: add Docker dev environment

PascalinDe:


lanterno (approved):
Nice work! the `dela.sh` adds more bash that we need to maintain, but I believe very soon we can refactor this, and remove a lot of the existing bash code, but for now let's merge this so we can get docker to work 🚀

dela.sh
```diff
@@ -0,0 +1,38 @@
+source .env;
```
pierluca, line 1: I'd suggest using a different filename to avoid confusion, especially as it's docker-specific.

docker-compose.yml
```diff
@@ -0,0 +1,81 @@
+services:
+
+  init-private-key:   # create private key
```
pierluca, line 3: isn't this creating a private key only to make it accessible to all the nodes? 🤔

dela.sh
pierluca, line 1: I'd also suggest adding a shebang and possibly a `bash -e` to ensure errors stop the script.
The script doesn't seem robust to failures.

pierluca (commented):
Minor comments :)

docker-compose.yml
PascalinDe: yes, that was not clear to me: as all I had to go on was a localhost example, all the nodes seem to be using the same private key

https://dedis.github.io/dela/#/memcoin

but I gather that this should not be the case?

dela.sh
PascalinDe: as Ahmed said, hopefully we can soon remove the script altogether it is mainly for our conveniance at the moment

nonetheless, I added the the shebang and removed the source

lanterno (approved):

docker-compose.yml
pierluca: Ah, I see where you're coming from.
The [BLS signature scheme](https://en.wikipedia.org/wiki/BLS_digital_signature) is used because it allows for signature aggregation. Effectively, each node can use its own signature and the resulting signature is an aggregation of the individual signatures. 

I think out of convenience the demo was using a single key, but separate keys might make sense as a proper example of how the system should work / be used.

pierluca (dismissed):

docker-compose.yml
PascalinDe: I'm currently on it, so please don't merge until after I've pushed the necessary changes :slightly_smiling_face:

pierluca (approved):

PR #290, opened 2023-08-02
style: renames variables, and applies linter

lanterno:


jbsv (approved):
I finally start to understand what it does :-)

PR #298, opened 2023-08-09
style: bump up project organization

lanterno:
This PR might be too opinionated, so feel free to comment everywhere, or reject the changes. 

But.. my goal with this PR was to 
1. Just upgrade everything to make sure we're following the trends (in a good way)
2. Reset config files to latest recommended versions (and omit the over configuration that we had)

I believe in minimal configuration that makes sense rather than having the best airbnb supported config that we don't fully utilise or understand.   
This should -maybe- minimise the project size, and makes it easier to understand/modify everything in the future. 

Updating everything to latest version as well should help us newbies as what we read in a library docs would match what we see in the code without doing much research! (it's also cool!)


PascalinDe:
+1 on the `.gitignore` I totally forgot about the `.env` file

no opinion on the rest though, I don't know the project enough yet to be able to decide this

web/backend/.eslintrc.cjs
pierluca, line 1: Isn't .cjs totally outdated as nearly everyone's moved away from CommonJS and towards ES Modules ? 😅

lanterno: the eslint documentation says 

> JavaScript (ESM) - use .eslintrc.cjs when running ESLint in JavaScript packages that specify "type":"module" in their package.json. Note that ESLint does not support ESM configuration at this time.

Maybe they just kept the legacy extension?

pierluca: what's weird is that normally ESM (ECMAScript Module) and CJS (CommonJS) are not the same thing, I don't know what to make of the documentation suggesting a cjs. 
I'd also consider changing the package.json to use `"type":"module"` instead of `"type": "commonjs"`, considering we're using the `import my_dependency` syntax and not `require("my_dependency")`

PR #289, opened 2023-08-02
Fix: update db utils script to reflect new architecture

lanterno:
The dbUtils script is outdated, and doesn't work with the current backend. I've adjusted the script, and tested it.

Currently, there's no tests, but it might be a good idea later to add some.

Reviewing each commit separately should provide a better idea of the logical sequence of changes.

ToDo:
- [x]  Rename the file to a more sensible name `cli.ts` 
- [x] Use a proper CLI library/tool
- [ ] Add some unit tests (maybe)

closes https://github.com/dedis/d-voting/issues/300

PR #299, opened 2023-08-09
fix: overhauled Docker environment

PascalinDe:
* cleaned up exposed ports
* added persistent storage to Dela nodes
* added reverse proxy to backend to avoid CORS issues


README.docker.md
```diff
@@ -52,6 +49,8 @@ to delete the volumes (this will reset your instance).
...
 ## Post-install commands
 
-1. run the script `DELA_REPLICAS=... init_dela.sh` to initialize the DELA network with `DELA_REPLICAS set to the same value as in .env`
-2. run `docker exec -it d-voting-backend-1 /bin/bash` to connect to the backend
-3. execute `node -e 'require("./dbUtils").addAdmin("./dvoting-users", <sciper>)'` with your Sciper to add yourself as admin
+1. `./init_dela.sh`
+2. `docker compose exec backend /bin/bash`
+3. `cd src`
+4. `node -e 'require("./dbUtils").addAdmin("<sciper>")'`
+5. exit the container and run `docker compose down && docker compose up -d`
```
lanterno, line 56: These changes will be affected by this PR https://github.com/dedis/d-voting/pull/289 if it got merged first

PascalinDe: I'll rebase the branch

docker-compose.yml
```diff
@@ -60,6 +60,7 @@ services:
       PRIVATE_KEY: ${PRIVATE_KEY}
     ports:
       - 127.0.0.1:5000:5000
+      - 127.0.0.1:80:80
```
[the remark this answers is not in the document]
PascalinDe: we need that because at some point the browser stops talking to the web app and talks directly to the Dela node URL (i.e. the proxy server deployed in the backend)

lanterno (dismissed):

docker-compose.yml.debug
pierluca, line 1: err, are you sure you don't want to name this `.debug.yml` so that the editors (including GitHub) don't freak out and still highlight the syntax correctly? 😁

init_dela.sh
pierluca, line 1: could this be run automatically as part of the initial `docker compose up` ?

Dockerfiles/nginx/default
```diff
@@ -0,0 +1,6 @@
+server {
+    listen 80;
```
pierluca, line 2: I would just include a note in the documentation that production deployments are expected to setup the necessary TLS stuff, in this container or elsewhere. That's really the (main) reason we advise to put nginx in front of the backend and the proxy.

pierluca (commented):
Looks pretty good to me, minor remarks.

Dockerfiles/nginx/default
PascalinDe: that Nginx shouldn't be there anyway, same reasoning as above: let's leave it like this for the moment and kick it later, when the problem is solved properly (i.e. CORS handled correctly)

I opened a ticket here: https://github.com/dedis/d-voting/issues/307

pierluca (approved):

PR #303, opened 2023-08-17
refactor: organise permissions into an object

lanterno:
I believe this new structure makes it a lot easier to read the permissions. 

when you read `PERMISSIONS.ACTIONS.LIST` this probably matches more how our brain would prefer the information to be presented.

### DIFF
--- a/web/backend/src/Server.ts
+++ b/web/backend/src/Server.ts
@@ -267,7 +267,7 @@ app.get('/api/user_rights', (req, res) => {
 // This call (only for admins) allow an admin to add a role to a voter.
 app.post('/api/add_role', (req, res, next) => {
   if (!isAuthorized(req.session.userid, PERMISSIONS.SUBJECTS.ROLES, PERMISSIONS.ACTIONS.ADD)) {
-    res.status(400).send({ error: 'user not logged-in or does not have enough permissions' });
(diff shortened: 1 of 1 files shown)


web/backend/src/Server.ts
```diff
@@ -259,8 +266,8 @@ app.get('/api/user_rights', (req, res) => {
 
 // This call (only for admins) allow an admin to add a role to a voter.
 app.post('/api/add_role', (req, res, next) => {
-  if (!isAuthorized(req.session.userid, SUBJECT_ROLES, ACTION_ADD)) {
-    res.status(400).send('Unauthorized - only admins allowed');
+  if (!isAuthorized(req.session.userid, PERMISSIONS.SUBJECTS.ROLES, PERMISSIONS.ACTIONS.ADD)) {
+    res.status(400).send({ error: 'user not logged-in or does not have enough permissions' });
```
pierluca, line 270: why return an `{ error: ...}` type here but not elsewhere?

pierluca (commented):
Thank you for this clean-up :)

web/backend/src/Server.ts
lanterno: ah! this slipped though my PR. I reverted it back. The next commit has a plan for my a better fix for this case.

PR #304, opened 2023-08-17
Refactor: makes session management a bit more readable

lanterno:
### Changes

1. Rename the session variables to follow camelCase.
2. Renames other various variables to provide more readability.
3. Return 401 when a user in unauthenticated, instead of 400.

for n. 3, I only apply that to one endpoint. In a following PR, I want apply this to all other endpoints, but I need to think of a smarter way to unify the behaviour between all endpoints that require authentication.


PascalinDe (approved):
looks great :)

pierluca:
Regarding #3 and your comment, you might want to look at ExpressJS error handling.
It might be possible to define a set of exceptions and just write a generic error handler based on the exception type: https://expressjs.com/en/guide/error-handling.html

web/frontend/src/index.tsx
```diff
@@ -224,38 +224,42 @@ const AppContainer = () => {
     };
 
     async function fetchData() {
+      const response = await fetch(ENDPOINT_PERSONAL_INFO, req);
+      let result;
       try {
```
pierluca, line 229: Beware ! This is a dangerous change.
The `fetch` API returns a promise.
`await`ing on a Promise can throw an exception if the Promise is rejected.
This needs to be in the try/catch block.

web/frontend/src/index.tsx
```diff
@@ -224,38 +224,42 @@ const AppContainer = () => {
...
+        } else if (response.status === 401) {
+          result = {
+            isLoggedIn: false,
+            firstName: '',
+            lastName: '',
+            authorization: {},
+          };
+        } else {
+          const txt = await response.text();
+          throw new Error(`unexpected magic status: ${response.status} - ${txt}`);
```
pierluca, line 241: why magic? just regular HTTP 😁 
It doesn't even have to be unexpected: it could also be an error when dealing with a reverse proxy in front of the backend.
It looks to me more like an unhandled response, which practically should indeed translate to some form of error.

web/frontend/src/index.tsx
```diff
@@ -224,38 +224,42 @@ const AppContainer = () => {
...
+          result = await response.json();
+        } else if (response.status === 401) {
+          result = {
+            isLoggedIn: false,
+            firstName: '',
+            lastName: '',
+            authorization: {},
+          };
+        } else {
+          const txt = await response.text();
```
pierluca, line 240: if the response did not return a 200, doesn't this always throw ?

web/frontend/src/index.tsx
```diff
@@ -224,38 +224,42 @@ const AppContainer = () => {
...
     async function fetchData() {
+      const response = await fetch(ENDPOINT_PERSONAL_INFO, req);
+      let result;
       try {
-        const res = await fetch(ENDPOINT_PERSONAL_INFO, req);
-
-        if (res.status !== 200) {
-          const txt = await res.text();
-          throw new Error(`unexpected status: ${res.status} - ${txt}`);
+        if (response.status === 200) {
```
pierluca, line 230: how do you feel about a switch statement ? 😁

web/frontend/src/index.tsx
PascalinDe: agree on the comment on the error message, it should be just the status code and the reason

web/frontend/src/index.tsx
lanterno: I changed it. Looks good 👍

web/frontend/src/index.tsx
lanterno: if it's not 200 or 401

web/frontend/src/index.tsx
lanterno: Yeah, I removed it  😁

web/frontend/src/index.tsx
lanterno: I changed this to put the try/catch around the whole function. What do you think?

pierluca:
Sonar seems to have a couple of good points

web/frontend/src/index.tsx
```diff
@@ -226,41 +226,48 @@ const AppContainer = () => {
...
+      // wait for the default proxy to be set
+      await setDefaultProxy();
+      setContent(<App />);
+    }
+    try {
+      fetchData();
+    } catch (e: any) {
+      setContent(<Failed>{e.toString()}</Failed>);
+      console.log('error:', e);
     }
```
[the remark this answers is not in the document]
lanterno: This already existed from the previous code, but that looks like a good improvement. I just applied it.

web/frontend/src/index.tsx
```diff
@@ -226,41 +226,48 @@ const AppContainer = () => {
...
+        },
+      });
+      // wait for the default proxy to be set
+      await setDefaultProxy();
+      setContent(<App />);
+    }
+    try {
+      fetchData();
+    } catch (e: any) {
+      setContent(<Failed>{e.toString()}</Failed>);
```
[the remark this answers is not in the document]
lanterno: I believe that is correct, but it already existed like this, so I wouldn't want to change it in this PR.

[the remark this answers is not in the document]
lanterno: But still I'm gonna take the advice for when I can apply am improvement confidently.

web/frontend/src/index.tsx
pierluca: what I meant is: isn't `response.text()` always going to throw before your `throw new Error()`, since we're not in a 200 ?

web/frontend/src/index.tsx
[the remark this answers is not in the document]
pierluca: Agreed, I'm just reviewing some of this code for the first time - so I'm pointing out issues as I see them. That doesn't mean I expect all the fixes in the same PR :)

pierluca (dismissed):
There are still a few issues, but nothing specifically belonging in this PR.
I think we can merge :)

pierluca (approved):

lanterno:
@pierluca Github dismissed your review for some reason.

web/frontend/src/index.tsx
lanterno: @pierluca It succeeds. I tested it. 

My test: I restarted the backend, and refreshed the UI causing the backend to return 504. the frontend then displayed the intended message.

pierluca (approved):

PR #313, opened 2023-08-29
feat: fix missing titles

PascalinDe:
The idea is to let `i18next` handle determining which language needs to be returned which also allows for setting a fallback in case no value has been set for that variable.

What remains to be done: set the English (or French?) title to be required

### DIFF
--- a/web/frontend/src/pages/form/components/FormRow.tsx
+++ b/web/frontend/src/pages/form/components/FormRow.tsx
@@ -15,21 +15,27 @@ const FormRow: FC<FormRowProps> = ({ form }) => {
   useEffect(() => {
     if (form.Title === '') return;
     if (isJson(form.Title)) {
-      const ts = JSON.parse(form.Title);
-      setTitles(ts);
+      setTitles(JSON.parse(form.Title));
     } else {
-      const t = { en: form.Title, fr: form.TitleFr, de: form.TitleDe };
-      setTitles(t);
+      setTitles({ en: form.Title, fr: form.TitleFr, de: form.TitleDe });
     }
   }, [form]);
+  // let i18next handle choosing the appropriate language
+  const formRowI18n = i18n.createInstance();
+  formRowI18n.init();
+  // get current language
+  formRowI18n.changeLanguage(i18n.language);
+  Object.entries(titles).forEach(([lang, title]: [string, string | undefined]) => {
+    if (title) {
+      formRowI18n.addResource(lang, 'form', 'title', title);
+    }
+  });
   return (
     <tr className="bg-white border-b hover:bg-gray-50 ">
       <td className="px-1.5 sm:px-6 py-4 font-medium text-gray-900 whitespace-nowrap truncate">
         <Link className="text-gray-700 hover:text-indigo-500" to={`/forms/${form.FormID}`}>
           <div className="max-w-[20vw] truncate">
-            {i18n.language === 'en' && titles.en}
-            {i18n.language === 'fr' && titles.fr}
-            {i18n.language === 'de' && titles.de}
+            {formRowI18n.t('title', { ns: 'form', fallbackLng: 'en' })}
           </div>
         </Link>
       </td>
--- a/web/frontend/src/schema/configurationValidation.ts
+++ b/web/frontend/src/schema/configurationValidation.ts
@@ -2,6 +2,11 @@ import * as yup from 'yup';
 
 const idSchema = yup.string().min(1).required();
 const titleSchema = yup.string().required();
+const formTitleSchema = yup.object({
+  en: yup.string().required(),
+  fr: yup.string(),
+  de: yup.string(),
+});
 
 const selectsSchema = yup.object({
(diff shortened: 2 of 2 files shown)


web/frontend/src/pages/form/components/FormRow.tsx
```diff
@@ -15,21 +15,25 @@ const FormRow: FC<FormRowProps> = ({ form }) => {
...
     }
   }, [form]);
+  const formRowI18n = i18n.createInstance();
+  formRowI18n.init();
+  formRowI18n.changeLanguage(i18n.language);
+  Object.entries(titles).forEach(([lang, title]: any[]) => {
+    if (title) {
+      formRowI18n.addResource(lang, 'form', 'title', title ?? '');
+    }
+  });
```
ineiti, line 30:
```suggestion
  Object.entries(titles).forEach(([lang, title]: [string, (string | undefined)]) => {
      formRowI18n.addResource(lang, 'form', 'title', title ?? "");
    });
```

However, this will be a bit different: now it even adds an entry for a language with an undefined `title`, but the `title` will be empty. If you want to avoid adding empty titles, replace the callback with this:

```
    if (title !== undefined){
      addResource(lang, 'form', 'title', title);
    }
```

ineiti: If that doesn't work, then you need to add a type definition to `titles`:

```
titles: {[key: string]: (string | undefined)}
```

Play around here:

https://www.typescriptlang.org/play?ssl=3&ssc=1&pln=4&pc=1#code/MYewdgzgLgBFCWUA2BTCAuGBvA2gaxQE9NoAneMAcwF1MAKMiymAHxgFcwATFAMwpRcAlAF8YAXmwAiAIZTMUgEZSANDCXyO3PgK4iA3AChDvTsATgYMrlwBKaEO1LAUdGSSjkqaxR6+U1YD8mNS5gqiEsQxgYmFBIEFQAOiQQSjdQoSMRYwB5RQArFHMklDBPeDQ6BGQ0ISTeEFIAURlgAAs6OhwkGW84RFRaGBxGfoYKqlYtHn4wQSFqIQkAPmxo2PheGGrBlBgAQnFJTlndSI3Yqxt7CEdnV17+gHJG0gBbZ7VnmtQvgdqWUuMByMREQKAA

ineiti: Yes, it would also be true. So you would need

```
if (title !== undefined && title !== "")
```

ineiti: Or you could also use

```
if (str?.length > 0)
```

web/frontend/src/pages/form/components/FormRow.tsx
```diff
@@ -15,21 +15,27 @@ const FormRow: FC<FormRowProps> = ({ form }) => {
...
+      setTitles({ en: form.Title, fr: form.TitleFr, de: form.TitleDe });
     }
   }, [form]);
+  // let i18next handle choosing the appropriate language
+  const formRowI18n = i18n.createInstance();
+  formRowI18n.init();
+  // get current language
+  formRowI18n.changeLanguage(i18n.language);
+  Object.entries(titles).forEach(([lang, title]: [string, string | undefined]) => {
+    if (title !== undefined && title !== '') {
```
ineiti, line 29:
```suggestion
    if (title?.length > 0) {
```

Sorry, last one...

ineiti (changes requested):

web/frontend/src/pages/form/components/FormRow.tsx
ineiti: Just for posteriority: sorry for the noise...

1. `(str !== undefined && str.length > 0)` is the most correct, but it's long
2. `(str?.length > 0)` potentially compares `undefined` and `0`, which is wrong and typescript will complain in strict mode - because I didn't read [StackOverflow](https://stackoverflow.com/a/3261380/1573347) correctly
3. `(str?.length)` would be better - now we're coercing either an `undefined` or a `number` into a `bool`
4. `(str)` coerces a `string` or an `undefined` into a boolean, which works, but shouldn't :)

I think I would go with 3., but 4. is also OK.

My preferred way would be:

```
Objects.entries(titles)
  .filter([_, title] => title?.length)
  .forEach([lang, title] => ...
```

But that's also wrong.

PascalinDe:
@pierluca this PR shows two small changes, but in fact the entire validation/internalisation code needs to be updated

however, I prefer to integrate small gradual changes that fix visible problems to

* avoid spending time on dead code (I'm pretty sure there's some)
* limit testing to one functionality (and making sure that works now)
* make quickly visible improvements instead of passing a lot of time on big changes

the idea being to move gradually in the right direction, is that OK for you?

Cc @ineiti @lanterno

ineiti:
sounds good. We'll discuss best next steps this Thursday...

ineiti (approved):

PR #348, opened 2023-09-08
Draft: use multi-staged build for clean frontend image

### LINKED ISSUE #342: Docker-image creation
When creating the `Dockerfiles/Dockerfile.frontend`, it should build the code and then only copy the `build/` directory (or whatever) into a clean docker image. This way the docker image doesn't start compiling the code when it is run.

PascalinDe:
closes #342

### DIFF
--- a/Dockerfiles/Dockerfile.frontend
+++ b/Dockerfiles/Dockerfile.frontend
@@ -1,8 +1,15 @@
-FROM node:20-bookworm
+FROM node:20-bookworm as build
 
 ENV REACT_APP_PROXY=http://localhost:9081
 ENV REACT_APP_NOMOCK=on
 WORKDIR /web/frontend
 COPY ../web/frontend .
-RUN npm install
-ENTRYPOINT ["npm", "start"]
+RUN npm ci
+RUN npm run build
+
+FROM node:20-bookworm as app
+
+WORKDIR /web/frontend
+COPY --from=build /web/frontend/build/ .
+
+ENTRYPOINT ["python3", "-m", "http.server", "3000", "--directory", "/web/frontend/", "--bind", "127.0.0.1"]


Dockerfiles/Dockerfile.frontend
```diff
@@ -1,8 +1,14 @@
...
 COPY ../web/frontend .
-RUN npm install
+RUN npm ci
+RUN npm run build
+
+FROM node:20-bookworm as app
+
+COPY --from=build /web/frontend/ .
+
 ENTRYPOINT ["npm", "start"]
```
ineiti, line 14: Does that work? You don't need `nodejs` anymore, as the `npm run build` will create just a set of static html/js/css files. So you can do:

https://github.com/c4dt/service-stainless/blob/main/webapp/Dockerfile#L11

Or 

https://github.com/c4dt/service-stainless/commit/bfeac27eb49bbd154246ff848c60c6400414390f#diff-06839e9bf05123a2ab7c9a5d721be661be9d231a6337bc7b8b0b6d8d8bf53d44R11

Or any other lightweight web server image.

ineiti (changes requested):
Did you try that? I'm surprised that `npm start` works without re-compiling everything.

Dockerfiles/Dockerfile.frontend
PascalinDe: I was testing it and it initially seemed fine, but now it's not working, so I put it back into Draft

before that it sort of complained that there was no package.json file as npm ci does not seem to generate one, so I copied everything over to test that

thanks for the comments, I think the error is in there somewhere (I was looking at the Dockerfiles just now, I think https://github.com/c4dt/service-spindle/blob/main/Dockerfile is probably the closest)

ineiti: > I was looking at the Dockerfiles

The spindle Dockerfile uses the `node` image, which is not necessary, as you should only have static files to serve for the frontend. So either a python or a webserver for static content is enough. Here is another, even more minimal webserver:

https://lipanski.com/posts/smallest-docker-image-static-website

PascalinDe:
the frontend UI seems to be working with this version now, but I'm having trouble receiving data from DELA now

PascalinDe:
OK, working now

Dockerfiles/Dockerfile.frontend
```diff
@@ -1,8 +1,15 @@
...
-ENTRYPOINT ["npm", "start"]
+RUN npm ci
+RUN npm run build
+
+FROM node:20-bookworm as app
+
+WORKDIR /web/frontend
+COPY --from=build /web/frontend/build/ .
+
+ENTRYPOINT ["python3", "-m", "http.server", "3000", "--directory", "/web/frontend/", "--bind", "127.0.0.1"]
```
lanterno, line 15: I see you removed `npm start` in favor of building the static site. There's one caveat to that, which is that during development, you will need to rebuild every time you do a change. 

With no `npm start`, you can get hot reloading, or at least you would need to do `docker compose restart container` 
With this design, you would need `docker .... build` first then `docker .... up` second.  

but... maybe,  if we change the entrypoint to a CMD, we can override it in compose file to a `npm run`

PascalinDe: please see @ineiti 's comments, now that we want to actually deploy it I agree that it makes sense to have a clean Docker image

personally, I have a separate Dockerfile for developing that kept the `npm start` for the reasons you mentioned

ineiti: You can also add a `Dockerfile.dev` with the `npm start` in it.

For the `npm start`, do you use the `-v` to link your local directory?

ineiti (approved):

Dockerfiles/Dockerfile.frontend
pierluca, line 15: why use python (and the `node` image) instead of a webserver, since that's the only thing we need? 🤔

PascalinDe:
the frontend server acts as a reverse proxy, so there needs to be a server running here as well - we'll shelve this PR for now

Issue #305, opened 2023-08-24
[Authorization] Missing permission for voting


lanterno:
The current version of the "web backend" service allows any user to vote as long as they are logged-in through EPFL's SSO "tequila". 

This doesn't match the requirements, as **each election will require a different set of users who are allowed to vote depending on the election.**

A new permission "Can vote" is needed on the object level. 
So a user A should have a "can vote" permission on form/election "X".

This will also modifications to the current Authorization framework as it's not currently configured to allow object-level permissions.

ToDo:
- [x] Add a "can vote" permission on an object level.  
- [x] Add ability to assign a user a "can vote" permission on a specific election/form from the CLI
- [x] Validate that logged-in user has "can vote" permission on an election/form before allowing the user to vote on an election/form.

Extra:
- [ ] Use the 3 roles **Admin**, **Operator**, **Voter** in the backend using **Casbin library**

Note: 
We use the [Casbin library](https://casbin.org/docs/overview/) to implement **Authorization**

lanterno:
Last item is still not done yet!

lanterno:
From your experience already, do you think this is completed now, or should we do something else?

PascalinDe:
> @ineiti @PascalinDe Do you think this is completed now, or should we do something else?

I'll close this, further improvements will be in new tickets

Issue #248, opened 2022-12-21
THREAT - A user can vote multiple times (count as multiple votes) in an election.


Flamewind97:
## Scenario
Whenever the frontend sends an encrypted ballot to the backend to sign, the backend will include the userID of the user in the ballot and sign the msg. This design will allow every user to vote at most once. However, since we are using Tequila as our login server, thus it is hard to create multiple valid users in Tequila during backend testing. A workaround was created. Backend implements a function called “makeID” to create a random ID and sign it along with the encrypted ballot. This will allow a user to vote multiple times which counts as multiple votes. And this function is not yet removed until now. Thus an adversary can vote as many times as he wants to manipulate the election results. 

## Where:
In the file “web/backend/Server.ts” function app.use(‘/api/evoting/*’)
```
app.use('/api/evoting/*', (req, res) => {
. . . 
// special case for voting
  const regex = /\/api\/evoting\/elections\/.*\/vote/;
  if (req.baseUrl.match(regex)) {
    // We must set the UserID to know who this ballot is associated to. This is
    // only needed to allow users to cast multiple ballots, where only the last
    // ballot is taken into account. To preserve anonymity the web-backend could
    // translate UserIDs to another random ID.
    // bodyData.UserID = req.session.userid.toString();
    bodyData.UserID = makeid(10);
  }
 
function makeid(length: number) {
  let result = '';
  const characters = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
  const charactersLength = characters.length;
  for (let i = 0; i < length; i += 1) {
    result += characters.charAt(Math.floor(Math.random() * charactersLength));
  }
  return result;
}
```

## Mitigation:
Remove the makeId function. If the developer prefers to keep the function for testing, we can set an env variable in config.env to determine this environment, which production environment would use User Id and the Dev environment would use Fake Id.

ineiti:
Perhaps we can put something like https://create-react-app.dev/docs/adding-custom-environment-variables/ in the code so it automatically chooses the debugging feature or not. Even though I did not completely understand whether `NODE_ENV` would serve our needs here.

pierluca:
I think indeed NODE_ENV could be the right approach here. That's what I've used in other projects and it makes things rather simple.

ineiti:
> I think indeed NODE_ENV could be the right approach here.

Currently we took the even simpler approach of adding a `console.warn("DEBUG code")`

PR #349, opened 2023-09-10
Feat: adds voting permission

lanterno:
This PR adds a new voting permission, and a CLI command that takes a list of voter IDs, and election ID to give those users voting permission over that election. 

closes https://github.com/dedis/d-voting/issues/305


web/backend/src/cli.ts
```diff
@@ -65,4 +67,28 @@ program
...
+  .action(async ({ electionId, scipersFile }) => {
+    fs.readFile(scipersFile, 'utf8', async (err: any, data: string) => {
+      if (err) {
+        console.error(err);
+        return;
+      }
+      const scipers: Array<string> = data.split('\n');
+      const policies = [];
+      for (let i = 0; i < scipers.length; i += 1) {
+        policies[i] = [scipers[i], electionId, PERMISSIONS.ACTIONS.VOTE];
```
ineiti, line 86: I would add a sanity-check here to make sure that the sciper is an actual number between 100000 and 999999, included. In case of failure, output the line number and the offending sciper.

web/backend/src/controllers/dela.ts
```diff
@@ -174,6 +176,29 @@ delaRouter.use('/services/shuffle/:formID', (req, res, next) => {
   }
   next();
 });
+
+delaRouter.post('/forms/:formID/vote', (req, res) => {
+  if (!req.session.userId) {
+    res.status(401).send('Authentication required!');
+    return;
+  }
+  if (!isAuthorized(req.session.userId, req.params.formID, PERMISSIONS.ACTIONS.VOTE)) {
```
ineiti, line 185: General question: is it possible for a user to spoof the `userId`, or is that some internal variable that is protected?

web/backend/src/controllers/dela.ts
```diff
@@ -174,6 +176,29 @@ delaRouter.use('/services/shuffle/:formID', (req, res, next) => {
...
+    return;
+  }
+
+  // We must set the UserID to know who this ballot is associated to. This is
+  // only needed to allow users to cast multiple ballots, where only the last
+  // ballot is taken into account. To preserve anonymity, the web-backend could
+  // translate UserIDs to another random ID.
+  // bodyData.UserID = req.session.userId.toString();
+  const bodyData = req.body;
+  bodyData.UserID = makeid(10);
```
ineiti, line 196: Add a comment that this is for debugging. In the best case even do some build-conditional code. Please have a look if the following example can do it:

https://a-carreras-c.medium.com/development-and-production-variables-for-react-apps-c04af8b430a5

```
// DEBUG: this is only for debugging and needs to be replaced before production
console.warn("Using debugging code to randomize userID")
```

ineiti (changes requested):

web/backend/src/controllers/dela.ts
lanterno: hmm, but why do you think this is only for debugging? (I actually didn't understand why this was done this way)

web/backend/src/controllers/dela.ts
pierluca: I don't remember - are the session cookies signed, and the signature verified ? (ExpressJS allows this, I'm not sure we use that option though)

If not, they can be spoofed.

web/backend/src/controllers/dela.ts
lanterno: I followed your suggestion, and added the comment, and the log. More work can be done when we look into the relevant ticket. (https://github.com/dedis/d-voting/issues/248)

web/backend/src/cli.ts
lanterno: I'm considering an addition like: 

```
if (scipers_list[i].length !== 6) {
      console.error(`Received a SCIPER of incorrect length on line: ${i + 1} - halting!`);
      return;
    }
```

However, I'm not sure it really adds any value. 

the CLI with file input is meant for system-to-system integration; the file is exported from EPFL systems, and imported here. (In fact, we still need the file format to be adjust the script accordingly). 

On the other hand, We don't have a particular reason to validate the operator's input as this data is external to our system.

ineiti: Wouldn't be the first time I use the wrong file, or that my file has some wrong data to it. Here it's admin-controlled data, and as an admin I would like to be informed if something is wrong with my file.

web/backend/src/controllers/dela.ts
ineiti: I'll open an issue.

web/backend/src/cli.ts
lanterno: > as an admin I would like to be informed if something is wrong with my file.

@ineiti  makes sense. I added a new commit with some validation.

web/backend/src/cli.ts
```diff
@@ -83,6 +83,17 @@ program
...
       const policies = [];
       for (let i = 0; i < scipers.length; i += 1) {
+        const sciper: number = Number(scipers[i]);
+        if (Number.isNaN(sciper)) {
+          console.error(`SCIPER on line ${i + 1} is not a number - exiting!`);
+          return;
+        }
+        if (sciper > 999999 || sciper < 100000) {
+          console.error(
+            `SCIPER on line ${i + 1} is outside acceptable range (100000 -> 999999) - exiting!`
```
ineiti, line 93:
```suggestion
            `SCIPER on line ${i + 1} is outside acceptable range (100000..999999) - exiting!`
```

web/backend/src/cli.ts
```diff
@@ -83,6 +83,17 @@ program
       const scipers: Array<string> = data.split('\n');
       const policies = [];
       for (let i = 0; i < scipers.length; i += 1) {
+        const sciper: number = Number(scipers[i]);
+        if (Number.isNaN(sciper)) {
+          console.error(`SCIPER on line ${i + 1} is not a number - exiting!`);
```
ineiti, line 88: This must throw an error:

https://www.npmjs.com/package/commander#custom-option-processing

So if you use it in a script, you get an actual different exit code.

Same thing for the sciper-number comparison.

ineiti (changes requested):

web/backend/src/cli.ts
ineiti: Added #356 , so closing this discussion.

web/backend/src/cli.ts
lanterno: changed.

web/backend/src/cli.ts
lanterno: Ok. now I'm raising an exception instead of printing and returning.

web/backend/src/cli.ts
```diff
@@ -65,4 +67,37 @@ program
...
+// the .voters.example file is available as an example
+program
+  .command('addVoters')
+  .description('Assigns a list of SCIPERs to an Election as Voters')
+  .requiredOption('-e, --election-id <char>', 'ID of the election')
+  .requiredOption('-sf, --scipers-file <char>', 'File with line-separated list of SCIPERs')
+  .action(async ({ electionId, scipersFile }) => {
+    fs.readFile(scipersFile, 'utf8', async (err: any, data: string) => {
+      if (err) {
+        console.error(err);
```
ineiti, line 80: Didn't see this one, which should also `throw`.

ineiti (changes requested):

web/backend/src/cli.ts
ineiti: Sorry, one more...

web/backend/src/cli.ts
lanterno: Yep, fixed as well.

web/backend/src/cli.ts
lanterno: done

ineiti (approved):

web/backend/src/cli.ts
```diff
@@ -65,4 +67,36 @@ program
...
+      const scipers: Array<string> = data.split('\n');
+      const policies = [];
+      for (let i = 0; i < scipers.length; i += 1) {
+        const sciper: number = Number(scipers[i]);
+        if (Number.isNaN(sciper)) {
+          throw new InvalidArgumentError(`SCIPER on line ${i + 1} is not a number - exiting!`);
+        }
+        if (sciper > 999999 || sciper < 100000) {
+          throw new InvalidArgumentError(
+            `SCIPER on line ${i + 1} is outside acceptable range (100000..999999) - exiting!`
```
pierluca, line 91: Having the line number is nice, but I'd consider outputting the SCIPER itself within quotes, so as to remove any ambiguity.

E.g.

```suggestion
          throw new InvalidArgumentError(`SCIPER '${sciper}' on line ${i + 1} is not a number - exiting!`);
        }
        if (sciper > 999999 || sciper < 100000) {
          throw new InvalidArgumentError(
            `SCIPER '${sciper}' on line ${i + 1} is outside acceptable range (100000..999999) - exiting!`
```

pierluca (approved):
Minor suggestion for an improvement, but looks good to me in general

web/backend/src/controllers/dela.ts
emduc: > hmm, but why do you think this is only for debugging? (I actually didn't understand why this was done this way)

It's very important to enforce one-person-one-vote. The current "makeId" randomizes the id so a user sending a vote will appear as a vote sent by a random user, thus you can re-vote as much as you want and duplicate your voting power.

pierluca (commented):
It seems lint is unhappy with your latest changes.

web/backend/src/controllers/dela.ts
```diff
@@ -174,6 +176,32 @@ delaRouter.use('/services/shuffle/:formID', (req, res, next) => {
...
+
+  // We must set the UserID to know who this ballot is associated to. This is
+  // only needed to allow users to cast multiple ballots, where only the last
+  // ballot is taken into account. To preserve anonymity, the web-backend could
+  // translate UserIDs to another random ID.
+  // bodyData.UserID = req.session.userId.toString();
+
+  // DEBUG: this is only for debugging and needs to be replaced before production
+  const bodyData = req.body;
+  console.warn('randomizing the SCIPER ID to allow for emulate votes - Use only for Development');
```
emduc, line 198: Personal preference but I would have tried a more explicit warning,  I have a bad tendency to ignore non-caps warnings.


```suggestion
  console.warn(' DEV CODE - randomizing the SCIPER ID to allow for unlimited votes - Use only for Development');
```

lanterno:
I will fix the linting, and rebase over main @ineiti @pierluca

web/backend/src/controllers/dela.ts
lanterno: done 👍

ineiti (changes requested):
Please re-run the failing go tests to make sure it's just a fluke.

lanterno:
@ineiti All tests are good now. 

I also looked at the go integration tests which fail mostly on the CI, and they don't depend on the Typescript code, so it's safe to ignore the failures in the future if there's no GoLang code changed.

ineiti (approved):

ineiti:
> @ineiti All tests are good now.

Great - for the other reviews of this PR I clicked myself on the "re-run tests" :)

PR #353, opened 2023-09-13
Adding keypair generation to cli.ts

ineiti:
Putting the keypair generation in typescript.

### DIFF
--- a/web/backend/readme.md
+++ b/web/backend/readme.md
@@ -10,16 +10,29 @@ Copy the database credentials from `config.env.template` to `/d-voting/bachend/s
 
 ## Generate a keypair
 
-Here is a small piece of code to help generating the keys:
+Here is a small piece of code in go to generate the keypair for the backend:
 
 ```go
-func GenerateKey() {
-	privK := suite.Scalar().Pick(suite.RandomStream())
-	pubK := suite.Point().Mul(privK, nil)
+package main
 
-  fmt.Println("PUBLIC_KEY:", pubK)
+import (
+	"fmt"
+	"go.dedis.ch/kyber/v3/group/edwards25519"
+	"go.dedis.ch/kyber/v3/util/key"
+)
+
+func main() {
+	pair := key.NewKeyPair(&edwards25519.SuiteEd25519{})
+
+    fmt.Println("PUBLIC_KEY:", pubK)
 	fmt.Println("PRIVATE_KEY:", privK)
 }
+```
+
+You can also use the `cli` program to generate the keys:
+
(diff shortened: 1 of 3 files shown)


web/backend/src/cli.ts
```diff
@@ -65,4 +66,15 @@ program
...
   });
 
+program
+  .command('keygen')
+  .description('Create a new keypair for the .env')
+  .action(() => {
+    const ed25519 = curve.newCurve('edwards25519');
+    const priv = ed25519.scalar().pick();
+    const pub = ed25519.point().mul(priv);
+    console.log(`PRIVATE_KEY=${priv}`);
```
emduc, line 76: is it normal to simply print them instead of saving them where they are needed?

ineiti: The goal here is to create the values that are needed for the `.env` file, so not to store them anywhere.

web/backend/readme.md
```diff
@@ -10,18 +10,31 @@ Copy the database credentials from `config.env.template` to `/d-voting/bachend/s
...
+	"fmt"
+	"go.dedis.ch/kyber/v3/group/edwards25519"
+	"go.dedis.ch/kyber/v3/util/key"
+)
+
+func main() {
+	pair := key.NewKeyPair(&edwards25519.SuiteEd25519{})
+
+    fmt.Println("PUBLIC_KEY:", pubK)
 	fmt.Println("PRIVATE_KEY:", privK)
```
pierluca, line 28: tabs vs. spaces, I'd suggest making this uniform :)

web/backend/readme.md
```diff
@@ -10,18 +10,31 @@ Copy the database credentials from `config.env.template` to `/d-voting/bachend/s
 
 ## Generate a keypair
 
-Here is a small piece of code to help generating the keys:
+Here is a small piece of code in go to generate the keypair for the backend:
```
pierluca, line 13:
```suggestion
Here is a small piece of code in Go to generate the keypair for the backend:
```

pierluca (commented):
Minor remarks.

web/backend/src/cli.ts
PascalinDe: maybe add a line saying "please make sure to store this somewhere" as people might indeed expect the values to also be written to a file

ineiti: Added an output, but now it won't be usable in scripts anymore...

emduc (approved):

pierluca (approved):

PR #343, opened 2023-09-06
Refactor: split code into modules

lanterno:
Splits the one-file project into different modules for different purposes. 

I think this is a good first iteration, and I will keep trying to improve the structure as I go.

You will also see that I refactored the authentication part a little bit as I moved it to a separate module `authManager.ts`. 

I believe all the module names will still evolve in the future, so no need to think about it too much right now.

PR #354, opened 2023-09-13
Fixing eslint warnings

ineiti:
Most of the warnings are trivial. I commented the ones I think are more complex.

### DIFF
--- a/web/frontend/src/mocks/handlers.ts
+++ b/web/frontend/src/mocks/handlers.ts
@@ -14,7 +14,6 @@ import {
   EditDKGActorBody,
   EditFormBody,
   NewDKGBody,
-  NewFormBody,
   NewFormVoteBody,
   NewProxyAddress,
   NewUserRole,
@@ -26,7 +25,6 @@ import { ID } from 'types/configuration';
 import { Action, Status } from 'types/form';
 import { setupMockForm, toLightFormInfo } from './setupMockForms';
 import setupMockUserDB from './setupMockUserDB';
-import { mockRoster } from './mockData';
 import { NodeStatus } from 'types/node';
 
 const uid = new ShortUniqueId({ length: 8 });
@@ -107,7 +105,7 @@ export const handlers = [
       })
     );
   }),
-  rest.put(endpoints.addFormAuthorization, async (req, res, ctx) => {
+  rest.put(endpoints.addFormAuthorization, async (req) => {
     const { FormID } = req.body as AddAuthBody;
     auth.set(FormID, ['own']);
   }),
@@ -123,36 +121,11 @@ export const handlers = [
   }),
 
   rest.post(endpoints.newForm, async (req, res, ctx) => {
-    const body = req.body as NewFormBody;
-
     await new Promise((r) => setTimeout(r, RESPONSE_TIME));
 
     if (!isAuthorized(auth, 'election', 'create')) {
       return res(ctx.status(403), ctx.json({ message: 'You are not authorized to create a form' }));
     }
-
-    const createForm = (configuration: any) => {
-      const newFormID = uid();
-      const newDKGStatus = new Map();
-      mockRoster.forEach((node) => {
-        newDKGStatus.set(node, NodeStatus.NotInitialized);
-      });
-      mockDKG.set(newFormID, newDKGStatus);
-
-      mockForms.set(newFormID, {
-        FormID: newFormID,
-        Status: Status.Initial,
-        Pubkey: 'DEAEV6EMII',
-        Result: [],
-        Roster: mockRoster,
-        Configuration: configuration,
-        BallotSize: 291,
-        ChunksPerBallot: 11,
-        Voters: [],
-      });
-
-      return newFormID;
-    };
 
     return res(ctx.status(200), ctx.json({ Status: 0, Token: fakeToken }));
   }),
@@ -460,7 +433,7 @@ export const handlers = [
     const body = req.body as UpdateProxyAddress;
     const node = decodeURIComponent(NodeAddr as string);
 
-    if (body.NewNode != node) {
+    if (body.NewNode !== node) {
       mockNodeProxyAddresses.delete(node);
       mockNodeProxyAddresses.set(body.NewNode, body.Proxy);
     } else {
--- a/web/frontend/src/pages/admin/Admin.tsx
+++ b/web/frontend/src/pages/admin/Admin.tsx
@@ -1,4 +1,4 @@
-import React, { FC, useContext, useEffect, useState } from 'react';
+import React, { FC, useContext, useEffect, useMemo, useState } from 'react';
 import { ENDPOINT_USER_RIGHTS } from 'components/utils/Endpoints';
 import { FlashContext, FlashLevel } from 'index';
 import Loading from 'pages/Loading';
@@ -14,7 +14,7 @@ const Admin: FC = () => {
   const [users, setUsers] = useState([]);
   const [loading, setLoading] = useState(true);
 
-  const abortController = new AbortController();
+  const abortController = useMemo(() => new AbortController(), []);
   const signal = abortController.signal;
 
   const request = {
@@ -52,7 +52,7 @@ const Admin: FC = () => {
     return () => {
       abortController.abort();
     };
-  }, [nodeProxyObject, nodeProxyError]);
+  }, [abortController, fctx, t, nodeProxyObject, nodeProxyError]);
 
   useEffect(() => {
     fetch(ENDPOINT_USER_RIGHTS)
--- a/web/frontend/src/pages/ballot/components/Rank.tsx
+++ b/web/frontend/src/pages/ballot/components/Rank.tsx
@@ -111,13 +111,14 @@ const Rank: FC<RankProps> = ({ rank, answers, language }) => {
               <ul className={rank.ID} {...provided.droppableProps} ref={provided.innerRef}>
                 {Array.from(answers.RankAnswers.get(rank.ID).entries()).map(
                   ([rankIndex, choiceIndex]) => {
-                    if (rank.ChoicesMap.get('en') === undefined) return;
-                    if (language === 'en' && rank.ChoicesMap.has('en'))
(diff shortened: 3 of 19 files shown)


web/frontend/src/mocks/handlers.ts
```diff
@@ -123,37 +121,12 @@ export const handlers = [
...
   rest.post(endpoints.newForm, async (req, res, ctx) => {
-    const body = req.body as NewFormBody;
-
     await new Promise((r) => setTimeout(r, RESPONSE_TIME));
 
     if (!isAuthorized(auth, 'election', 'create')) {
       return res(ctx.status(403), ctx.json({ message: 'You are not authorized to create a form' }));
     }
 
-    const createForm = (configuration: any) => {
```
ineiti, line 134: `createForm` is never called, as such I think it's safe to remove it.

web/frontend/src/pages/admin/Admin.tsx
```diff
@@ -14,7 +14,7 @@ const Admin: FC = () => {
   const [users, setUsers] = useState([]);
   const [loading, setLoading] = useState(true);
 
-  const abortController = new AbortController();
+  const abortController = useMemo(() => new AbortController(), []);
```
ineiti, line 17: See https://stackoverflow.com/questions/69572613/how-to-wrap-the-initialisation-of-an-array-in-its-own-usememo-hook

web/frontend/src/pages/ballot/components/Rank.tsx
```diff
@@ -111,13 +111,14 @@ const Rank: FC<RankProps> = ({ rank, answers, language }) => {
...
                   ([rankIndex, choiceIndex]) => {
-                    if (rank.ChoicesMap.get('en') === undefined) return;
-                    if (language === 'en' && rank.ChoicesMap.has('en'))
-                      return choiceDisplay(rank.ChoicesMap.get('en')[choiceIndex], rankIndex);
-                    else if (language === 'fr' && rank.ChoicesMap.has('fr'))
+                    if (language === 'fr' && rank.ChoicesMap.has('fr'))
                       return choiceDisplay(rank.ChoicesMap.get('fr')[choiceIndex], rankIndex);
                     else if (language === 'de' && rank.ChoicesMap.has('de'))
                       return choiceDisplay(rank.ChoicesMap.get('de')[choiceIndex], rankIndex);
+                    else if (rank.ChoicesMap.has('en'))
```
ineiti, line 118: As `en` is the default language, I moved it to the end of the `if` clauses and removed the `language` comparison. So if it's neither `de`, nor `fr`, or if the chosen language is not available, it tries to use `en`.

web/frontend/src/pages/form/IndividualResult.tsx
```diff
@@ -62,82 +62,89 @@ const IndividualResult: FC<IndividualResultProps> = ({
     OUT_OF_BOUNDS = 2,
   }
 
-  const questionIcons = {
-    [RANK]: <SwitchVerticalIcon />,
-    [SELECT]: <CursorClickIcon />,
-    [TEXT]: <MenuAlt1Icon />,
-  };
-  const SubjectElementResultDisplay = (element: SubjectElement) => {
```
ineiti, line 70: See https://dev.to/amitaldo/function-makes-the-dependencies-of-useeffect-hook-change-on-every-render-warning-in-react-35hd

web/frontend/src/pages/form/IndividualResult.tsx
```diff
@@ -62,82 +62,89 @@ const IndividualResult: FC<IndividualResultProps> = ({
...
-              </div>
-            ) : (
-              SubjectElementResultDisplay(subject.Elements.get(id))
-            )}
-          </div>
-        ))}
-      </div>
-    );
-  };
+  const displayResults = useCallback(
```
ineiti, line 116: See https://dev.to/amitaldo/function-makes-the-dependencies-of-useeffect-hook-change-on-every-render-warning-in-react-35hd

web/frontend/src/pages/form/components/utils/useQuestionForm.ts
```diff
@@ -142,9 +142,7 @@ const useQuestionForm = (initState: RankQuestion | SelectQuestion | TextQuestion
         const newChoicesMapDefault = new Map(
           Object.entries({
             ...obj,
-            ['en']: obj.en.map((item: string, idx: number) =>
-              idx === index ? e.target.value : item
-            ),
+            en: obj.en.map((item: string, idx: number) => (idx === index ? e.target.value : item)),
```
pierluca, line 145: Very subjective but I think it was more readable before 😅 
Furthermore, we try to (roughly) keep a 80 char per line limit.
I haven't checked, but I suspect this is longer than that.

web/frontend/src/pages/form/IndividualResult.tsx
```diff
@@ -62,82 +62,89 @@ const IndividualResult: FC<IndividualResultProps> = ({
...
-    }
-    if (titles === undefined) {
-      titles = { en: element.Title, fr: element.TitleFr, de: element.TitleDe };
-    }
-    return (
-      <div className="pl-4 pb-4 sm:pl-6 sm:pb-6">
-        <div className="flex flex-row">
-          <div className="align-text-middle flex mt-1 mr-2 h-5 w-5" aria-hidden="true">
-            {questionIcons[element.Type]}
+  const SubjectElementResultDisplay = useCallback(
```
pierluca, line 65: It looks like, instead of having a callback, this could/should be its own component.

web/frontend/src/pages/form/IndividualResult.tsx
pierluca: same, this looks like it should be moved to its own component, maybe combined with the one above.

web/frontend/src/pages/ballot/components/Select.tsx
```diff
@@ -102,25 +102,25 @@ const Select: FC<SelectProps> = ({ select, answers, setAnswers, language }) => {
...
               );
-            else if (language === 'de' && select.ChoicesMap.has('de'))
+            else if (select.ChoicesMap.has('en'))
               return choiceDisplay(
                 isChecked,
-                select.ChoicesMap.get('de')[choiceIndex],
+                select.ChoicesMap.get('en')[choiceIndex],
                 choiceIndex
               );
+            return '';
```
pierluca, line 123: Why not just `return` as it was the case before ?

web/frontend/src/pages/ballot/components/Rank.tsx
pierluca: The default `return` (undefined) was replaced by an empty string though. I'm not sure if that's desirable in `tsx` syntax: undefined clearly means there should be no output.

pierluca (commented):
Bunch of tiny comments :-)

I'm a little more worried about the Sonar security warnings though, I thought I'd find them addressed here, but in hindsight they probably deserve their own PR.

web/frontend/src/pages/ballot/components/Select.tsx
ineiti: Because eslint complains that `map` wants a value to be returned.

web/frontend/src/pages/ballot/components/Rank.tsx
ineiti: eslint complained that `map` wants to have a value to return. If you think `undefined` is the correct return here, I'll `return undefined`.

ineiti:
> I'm a little more worried about the Sonar security warnings though

Once #343 goes in, I'll have a look at those.

web/frontend/src/pages/form/components/utils/useQuestionForm.ts
ineiti: Following eslint warning. The `,` is on position 99, so it should just fit :)

What I don't understand is that eslint's [max-len](https://eslint.org/docs/latest/rules/max-len) should cut it off at 80 - should I add a `max-len` entry to `web/frontend/.eslintrc`?

web/frontend/src/pages/form/IndividualResult.tsx
ineiti: Long time no React here.

Do you mean putting `SubjectElementResultDisplay` and `displayResults` in a separate file? Or something like https://www.w3schools.com/react/react_components.asp ?

web/frontend/src/pages/ballot/components/Rank.tsx
ineiti: Github bug: I saw your comment twice, even after reloading the page. Now that I deleted the comment once, it's gone twice. Go figure - sorry...

web/frontend/src/pages/form/IndividualResult.tsx
ineiti: As I was just fixing eslint warnings, I propose to leave it as-is. Same for `displayResults`.

web/frontend/src/pages/ballot/components/Rank.tsx
pierluca: Thank you for interpreting my comment correctly - yes for filtering ! 😁

web/frontend/src/pages/form/IndividualResult.tsx
pierluca: I meant its own file + its own exported function.
These days React components are mostly functions, no need for a class
https://react.dev/learn/your-first-component

pierluca (approved):
As per your comment, feel free to leave things as-is if you're not keen to remove complexity at this point.

web/frontend/src/pages/form/IndividualResult.tsx
ineiti: Here is where I don't know what to do: `displayResults` uses `SubjectElementResultDisplay`, which takes four parameters from the `IndividualResult` component. So I'm not sure how to turn my head to tell React when it should redraw, and how to write the `displayResults` and `SubjectElementResultDisplay` to make it all good.

As this was mostly a linting PR, I'll merge as-is.

PR #355, opened 2023-09-13
Renaming memcoin to dvoting

ineiti:
Making sure the name of the binary makes sense.


emduc (approved):
Makes a lot of sense if it doesn't conflict with existing files

PR #366, opened 2023-09-25
Adds local login and script for testing

ineiti:
- Allow for easier development with bypassing the Gaspar login
- Adds a script which does all the initialisations for a first D-voting experience

### DIFF
--- a/web/backend/src/controllers/authentication.ts
+++ b/web/backend/src/controllers/authentication.ts
@@ -5,6 +5,37 @@ import { getUserPermissions, setMapAuthorization } from '../authManager';
 
 export const authenticationRouter = express.Router();
 
+authenticationRouter.get('/get_dev_login', (req, res) => {
+  if (process.env.NODE_ENV !== 'development') {
+    const err = `/get_dev_login can only be called in development: ${process.env.NODE_ENV}`;
+    console.error(err);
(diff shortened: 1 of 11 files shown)


web/backend/src/controllers/authentication.ts
```diff
@@ -5,6 +5,37 @@ import { getUserPermissions, setMapAuthorization } from '../authManager';
...
+    res.status(500).send(err);
+    return;
+  }
+  const sciper = parseInt(process.env.SCIPER_ADMIN, 10);
+  if (sciper < 100000 || sciper > 999999) {
+    const err = 'SCIPER_ADMIN must be between 100000 and 999999 included';
+    console.error(err);
+    res.status(500).send(err);
+    return;
+  }
```
pierluca, line 27: since I've seen similar checks pop up a bunch of times, are we sure we don't want a `Sciper` class that inherits from `Number` and implements a `AssertValid` function that throws errors ? 😄

ineiti: A class seemed to be too much, because it's mostly used as a number afterwards. But I added a method for the three invocations of this test.

ineiti:
Closed in favour of c4dt/d-voting#4

PR #367, opened 2023-09-25
Run voting tests

ineiti:


ineiti:
Closed in favour if c4dt/d-voting#2

PR #368, opened 2023-09-25
Consistent sleeps throughout all tests

emduc:
Adds a simple sleep after form creation in 1 test where it was missing and usually fails the pipeline.


pierluca (approved):
I'm pretty sure this is not the most robust fix ever, but it beats failing tests... let's hope we can devise a better waiting mechanism in the future 😅

PR #370, opened 2023-09-25
Using smallest webserver

ineiti:


pierluca (approved):
This looks nice, just wondering: what's busybox like, security-wise ?
I have never used it outside cheapo router firmwares.
I assume you know more than me about it :-)

ineiti:
Two things:
- it should probably not be used in production
- as the d-voting frontend does proxying to the backend and to the dela-proxies, this won't work anyway

PR #371, opened 2023-09-28
Fixing voting

ineiti:
Using the endpoint 'forms/:formID/vote' doesn't work with the sendToDela method

### DIFF
--- a/web/backend/src/controllers/dela.ts
+++ b/web/backend/src/controllers/dela.ts
@@ -177,31 +177,6 @@ delaRouter.use('/services/shuffle/:formID', (req, res, next) => {
   next();
 });
 
-delaRouter.post('/forms/:formID/vote', (req, res) => {
-  if (!req.session.userId) {
-    res.status(401).send('Authentication required!');
-    return;
-  }
-  if (!isAuthorized(req.session.userId, req.params.formID, PERMISSIONS.ACTIONS.VOTE)) {
-    res.status(400).send('Unauthorized');
-    return;
-  }
-
-  // We must set the UserID to know who this ballot is associated to. This is
-  // only needed to allow users to cast multiple ballots, where only the last
-  // ballot is taken into account. To preserve anonymity, the web-backend could
-  // translate UserIDs to another random ID.
-  // bodyData.UserID = req.session.userId.toString();
-
-  // DEBUG: this is only for debugging and needs to be replaced before production
-  const bodyData = req.body;
-  console.warn('DEV CODE - randomizing the SCIPER ID to allow for unlimited votes');
-  bodyData.UserID = makeid(10);
-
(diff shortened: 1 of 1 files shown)


web/backend/src/controllers/dela.ts
```diff
@@ -263,6 +238,30 @@ delaRouter.use('/*', (req, res) => {
...
 
   const bodyData = req.body;
+
+  // special case for voting
+  const match = req.baseUrl.match('/api/evoting/forms/(.*)/vote');
+  if (match) {
+    if (!req.session.userId) {
+      res.status(401).send('Authentication required!');
+      return;
+    }
```
pierluca, line 248: this will never be called based on the code in line 235-237. 
That error should probably return a 401, like you've done here.

web/backend/src/controllers/dela.ts
```diff
@@ -263,6 +238,30 @@ delaRouter.use('/*', (req, res) => {
   }
 
   const bodyData = req.body;
+
+  // special case for voting
+  const match = req.baseUrl.match('/api/evoting/forms/(.*)/vote');
```
pierluca, line 243: why not set this up as a specific endpoint ? It would be more idiomatic, unless I'm missing something.

web/backend/src/controllers/dela.ts
```diff
@@ -263,6 +238,30 @@ delaRouter.use('/*', (req, res) => {
...
+
+    // We must set the UserID to know who this ballot is associated to. This is
+    // only needed to allow users to cast multiple ballots, where only the last
+    // ballot is taken into account. To preserve anonymity, the web-backend could
+    // translate UserIDs to another random ID.
+    // bodyData.UserID = req.session.userId.toString();
+
+    // DEBUG: this is only for debugging and needs to be replaced before production
+    console.warn('DEV CODE - randomizing the SCIPER ID to allow for unlimited votes');
+    bodyData.UserID = makeid(10);
```
pierluca, line 262: can we use a global env. variable to define if we're in debug or not, instead of commenting out ? I'm afraid something like this will make its way back into production someday when someone's looking in the other direction :-)

web/backend/src/controllers/dela.ts
ineiti: That's what we did, but then `sendToDela` fails, because it uses `baseUrl`, which is different in `delaRouter.Post` and `delaRouter.use`.

web/backend/src/controllers/dela.ts
ineiti: Definitely better to use a global variable. One of the problems is that we cannot use `NODE_ENV === 'development'`, as we currently need to run the frontend in development mode due to the proxying.

ineiti:
Closing this in favour of https://github.com/c4dt/d-voting/pull/20

PR #396, opened 2025-03-26
revert go.dedis.ch to github.com/dedis

jbsv:
Reverting the paths to github.com/dedis allows the integration tests to pass.
This issue of go import will be investigated further. For the time being, the code base is all tests passing.


ineiti:
May I ask why the change back?

jbsv:
Using the paths to go.dedis.ch/d-voting breaks the integration test. It is very unclear why for now. But I prefer to leave a working set of tests in the code base.

